PR #79, opened 2020-09-28
Logger implmentation

HansikaPH:


ourownstory (commented):
I fixed the logger to be implemented in a more general way, allowing for hierarchical verbosity and traceability.

PR #80, opened 2020-09-29
update name of forecast_in_focus to highlight_forecast_step_n

ourownstory:
update name of forecast_in_focus to highlight_forecast_step_n


ourownstory:
@HansikaPH please note name change.

PR #127, opened 2020-11-29
Make docs

HansikaPH:


ourownstory (approved):
Looks good!
in our next iteration, we should find a way to make it less text-heavy and better maintainable (code changes should either automatically be reflected in the documentation or break it)

PR #289, opened 2021-03-03
Early Stopping

fjosp:
Add functionality that allows early stopping based on smoothed loss.


karl-richter:
Closing with this pr since early stopping will be introduced with the Lightning migration #837.

PR #299, opened 2021-03-24
Manik tharaka add plotly capabilities

ourownstory:


ourownstory:
Hi @manikTharaka great work!
Hansika and I tried out the plotly function and got it to work with both interactive plotting and saving the plot. That's wonderful!

There are still a few points that we should address before merging:
- create complete tests for all plotting functions  (I already added a simple example test for the main plotly function (not complete))
- test if an install from scratch with the (added) requirements in a new venv works without errors (I also added requirement for figure saving, kaleido, see requirements). The user should not have to install anything manually.
- add a small example notebook showing how to use the plotting functions, incl. saving figures.
- optional: add plot_parameters functionality

manikTharaka:
Hi @ourownstory ,

Noted, will make the necessary updates. Small concern regarding the second point above, I think in the case of Prophet it does not install the Plotly libraries during installation. I believe it is more of a soft dependency and is expected that the user will install Plotly if they need the use of Plotly as stated in the section about Plotly [here](https://facebook.github.io/prophet/docs/quick_start.html#python-api). Just wanted to point it out in case if we need to adopt a similar practice.

ourownstory:
> expected that the user will install Plotly if they need the use of Plotly
Hi @manikTharaka 
That is a good point. 
In our case, we do aim to eventually deprecate matplotlib and completely replace it with plotly, unless we run into major issues. This will make the code maintenance simpler, as we only have one plotting framework.
Thus, I think it makes sense to include it by default.

ourownstory:
continued in #322

PR #321, opened 2021-04-30
Energy datasets

GonzagueHenri:
to specify where the data is from (according to their license)


GonzagueHenri:
I added some energy data and a readme to cite the source

ourownstory:
Thank you @GonzagueHenri !

PR #358, opened 2021-07-15
Allow missing values left unimputed - drop affected areas

sobyah:
Implementing a fix for Issue#52:

- Allow missing values for Autoregression when opting not to impute
- After creating time dataset, remove the windows that have ANY missing values


ourownstory:
merged in #619

PR #322, opened 2021-05-10
Manik tharaka add plotly capabilities

manikTharaka:
Added tests and plot parameters functionality

PR #255, opened 2021-02-03
Uncertainty: Quantile Regression

HansikaPH:

PR #368, opened 2021-07-21
Plotly: support plotting of quantiles 

lsy44415:
Extend plot, plot component and plot parameter implementation from: 
- Manik tharaka add plotly capabilities #322 
- Overall prediction intervals #255. 

Port Matplotlib to Plotly and add overall prediction quantiles intervals in plot funtion.


ourownstory:
will be merged in #726

PR #384, opened 2021-08-28
n_regressors added to NP

MateusGheorghe:
n_regressors allows the use of covariates with independent lags from n_lags. One can now use covariates with auto-regression off.


ourownstory:
@MateusGheorghe 
I had a quick look at the changes, and it looks mostly good. I have a few questions:
* Can we store n_regressor inside the covar config (adding an extra variable "n_lags" to it's dataclass)?
* Why rename the arg to n_ar, but not rename the self.n_lags variable storing it? 
* What does allow_nnet_covar do? It seems like it is used for multiple things in the code?

MateusGheorghe:
> Can we store n_regressor inside the covar config (adding an extra variable "n_lags" to it's dataclass)?
Not sure if I understood what you mean. Do you mean n_regressors to the cover config?

> Why rename the arg to n_ar, but not rename the self.n_lags variable storing it?
I actually never renamed it n_ar, I thought about it and then decided to leave it as n_lags (since it would require changing the variable in many places. If you notice the only time n_ar shows up is in the forecaster line 99 - "n_ar (int): Previous time series steps to include in auto-regression. Aka AR-order" which I left in the code by accident since it was only a comment.

> What does allow_nnet_covar do? It seems like it is used for multiple things in the code?
It is a flag that is true when n_lags=0 and n_regressors>0. It prevents that the model is fit such as in the Prophet because n_lags=0 (but n_regressors >0).

neuralprophet/time_dataset.py
```diff
@@ -193,30 +195,32 @@ def _stride_time_features_for_forecasts(x):
...
         for name, features in seasonalities.items():
-            if n_lags == 0:
+            if aux_lags == 0:
                 seasonalities[name] = np.expand_dims(features, axis=1)
             else:
                 # stride into num_forecast at dim=1 for each sample, just like we did with time
                 seasonalities[name] = _stride_time_features_for_forecasts(features)
         inputs["seasonalities"] = seasonalities
 
     def _stride_lagged_features(df_col_name, feature_dims):
```
ourownstory, line 205: assert feature_dims >= 1

neuralprophet/forecaster.py
```diff
@@ -382,20 +378,21 @@ def add_lagged_regressor(self, names, regularization=None, normalize="auto", onl
                     * (default) ``False`` use same number of lags as auto-regression
                     * ``True`` only use last known value as input
         """
+        if n_covars == "auto":
+            n_covars = self.n_lags
```
ourownstory, line 382: if self.N-lags is not None and self.N-lags > 0:
= n_lags
else
= 1

neuralprophet/forecaster.py
```diff
@@ -361,7 +355,9 @@ def __init__(
         self.highlight_forecast_step_n = None
         self.true_ar_weights = None
 
-    def add_lagged_regressor(self, names, regularization=None, normalize="auto", only_last_value=False):
+    def add_lagged_regressor(
+        self, names, n_covars="auto", regularization=None, normalize="auto", only_last_value=False
```
ourownstory, line 359: drop kwarg only_last_value

neuralprophet/forecaster.py
```diff
@@ -382,20 +378,21 @@ def add_lagged_regressor(self, names, regularization=None, normalize="auto", onl
...
             names = [names]
         for name in names:
             self._validate_column_name(name)
             if self.config_covar is None:
                 self.config_covar = OrderedDict({})
             self.config_covar[name] = configure.Covar(
-                reg_lambda=regularization,
-                normalize=normalize,
-                as_scalar=only_last_value,
+                reg_lambda=regularization, normalize=normalize, as_scalar=only_last_value, n_covars=n_covars
```
ourownstory, line 395: as_scalar = n_covars == 1

neuralprophet/time_net.py
```diff
@@ -159,11 +160,11 @@ def __init__(
         # Covariates
         self.config_covar = config_covar
         if self.config_covar is not None:
-            assert self.n_lags > 0
+            assert check_n_lags_and_n_covars(self.config_covar, n_lags) > 0
```
ourownstory, line 163: can drop assert

neuralprophet/time_net.py
```diff
@@ -159,11 +160,11 @@ def __init__(
...
             self.covar_nets = nn.ModuleDict({})
             for covar in self.config_covar.keys():
                 covar_net = nn.ModuleList()
-                d_inputs = self.n_lags
+                d_inputs = self.config_covar[covar].n_covars
                 if self.config_covar[covar].as_scalar:
```
ourownstory, line 168: drop this, n_covars is already =1 in this case

neuralprophet/time_dataset.py
```diff
@@ -193,30 +195,32 @@ def _stride_time_features_for_forecasts(x):
...
-    if covar_config is not None and n_lags > 0:
+    if covar_config is not None and aux_lags > 0:
         covariates = OrderedDict({})
         for covar in df.columns:
             if covar in covar_config:
-                assert n_lags > 0
-                window = n_lags
+                assert covar_config[covar].n_covars > 0
+                window = covar_config[covar].n_covars
                 if covar_config[covar].as_scalar:
```
ourownstory, line 224: drop this, n_covars is already =1 in this case

tests/test_integration.py
```diff
@@ -1207,3 +1207,43 @@ def test_progress_display():
...
         )
         metrics_df = m.fit(df, progress=progress)
+
+
+def test_n_covars():
+    df = pd.read_csv(PEYTON_FILE, nrows=NROWS)
+    df1 = df.iloc[:128, :].copy(deep=True)
+    df1["A"] = df1["y"].rolling(30, min_periods=1).mean()
+    n_lags_input = [2, 2, 5, 2, 0]
+    n_covars_input = [2, "auto", 2, 5, 5]
```
ourownstory, line 1217: add 1

tests/test_integration.py
```diff
@@ -1207,3 +1207,43 @@ def test_progress_display():
             learning_rate=LR,
         )
         metrics_df = m.fit(df, progress=progress)
+
+
+def test_n_covars():
+    df = pd.read_csv(PEYTON_FILE, nrows=NROWS)
+    df1 = df.iloc[:128, :].copy(deep=True)
+    df1["A"] = df1["y"].rolling(30, min_periods=1).mean()
+    n_lags_input = [2, 2, 5, 2, 0]
```
ourownstory, line 1216: replace one value with 1

tests/test_integration.py
```diff
@@ -1207,3 +1207,43 @@ def test_progress_display():
...
+        log.debug(info_input[i])
+        m = NeuralProphet(n_forecasts=2, n_lags=n_lags_input[i], epochs=EPOCHS, batch_size=BATCH_SIZE, learning_rate=LR)
+        m = m.add_lagged_regressor(names="A", n_covars=n_covars_input[i])
+        metrics = m.fit(df1, freq="D")
+        future = m.make_future_dataframe(df1, n_historic_predictions=True)
+        forecast = m.predict(df=future)
+        if PLOT:
+            fig = m.plot(forecast)
+            fig = m.plot_parameters()
+    df1["B"] = df1["y"].rolling(8, min_periods=1).mean()
```
ourownstory, line 1235: add comment about purpose of test, maybe make a separate test function for the rest?

ourownstory (commented):
also add self.max_lags

neuralprophet/configure.py
```diff
@@ -310,6 +310,7 @@ class Covar:
     reg_lambda: float
     as_scalar: bool
     normalize: (bool, str)
+    n_covars: int
```
ourownstory, line 313: is this N-lags?

neuralprophet/df_utils.py
```diff
@@ -90,6 +90,34 @@ def join_dataframes(df_dict):
     return df_joined, episodes
 
 
+def check_n_lags_and_n_covars(config_covar, n_lags):
```
ourownstory, line 93: you could make it clearer what the function does with the name, e.g. `get_max_num_lags`

ourownstory: you could also simplify the function to sth like:
return max([n_lags] + [val.n_covars for key, val in config_covar.items()]

neuralprophet/forecaster.py
```diff
@@ -328,12 +328,7 @@ def __init__(
         # AR
         self.config_ar = configure.from_kwargs(configure.AR, kwargs)
         self.n_lags = self.config_ar.n_lags
-        if n_lags == 0 and n_forecasts > 1:
-            self.n_forecasts = 1
-            log.warning(
-                "Changing n_forecasts to 1. Without lags, the forecast can be "
-                "computed for any future time, independent of lagged values"
-            )
+        self.max_lags = self.n_lags
```
ourownstory, line 331: should this not also check for covars?

neuralprophet/forecaster.py
```diff
@@ -375,7 +370,13 @@ def __init__(
         self.highlight_forecast_step_n = None
         self.true_ar_weights = None
 
-    def add_lagged_regressor(self, names, regularization=None, normalize="auto", only_last_value=False):
+    def add_lagged_regressor(
+        self,
+        names,
+        n_covars="auto",
```
ourownstory, line 376: n_lags may be a clearer arg name

neuralprophet/forecaster.py
```diff
@@ -384,32 +385,39 @@ def add_lagged_regressor(self, names, regularization=None, normalize="auto", onl
...
-                specifies last value handling
-
-                Options
-                    * (default) ``False`` use same number of lags as auto-regression
-                    * ``True`` only use last known value as input
         """
+        if n_covars == "auto":
+            if self.n_lags is not None and self.n_lags > 0:
+                n_covars = self.n_lags
+                log.info("n_covars is equal to n_lags")
```
ourownstory, line 401: could make log message a little clearer

neuralprophet/forecaster.py
```diff
@@ -384,32 +385,39 @@ def add_lagged_regressor(self, names, regularization=None, normalize="auto", onl
...
+            else:
+                n_covars = 1
+                log.info("n_covars is equal to 1")
```
ourownstory, line 404: could make log message a little clearer

neuralprophet/forecaster.py
```diff
@@ -384,32 +385,39 @@ def add_lagged_regressor(self, names, regularization=None, normalize="auto", onl
...
+        if n_covars == "scalar":
+            n_covars = 1
+            log.info("n_covars is equal to 1")
```
ourownstory, line 407: could make log message a little clearer

neuralprophet/forecaster.py
```diff
@@ -384,32 +385,39 @@ def add_lagged_regressor(self, names, regularization=None, normalize="auto", onl
...
+        only_last_value = False if n_covars > 1 else True
         if self.fitted:
             raise Exception("Covariates must be added prior to model fitting.")
-        if self.n_lags == 0:
-            raise Exception("Covariates must be set jointly with Auto-Regression.")
+        if n_covars == 0 or n_covars is None:
+            raise Exception("Please, set number of lags for covariates (n_covars > 0)")
```
ourownstory, line 412: We could instead log a warning, and just do nothing (return self) -> this would not be a fatal error

neuralprophet/time_net.py
```diff
@@ -170,7 +170,7 @@ def __init__(
             self.config_events = None
             self.config_holidays = None
 
-            # Autoregression
+        # Autoregression
         self.n_lags = n_lags
         self.num_hidden_layers = num_hidden_layers
         self.d_hidden = n_lags + n_forecasts if d_hidden is None else d_hidden
```
ourownstory, line 176: let's make this:
if d_hidden is None:
    self.d_hidden = max(4, (n_lags + n_forecasts) /(2.0* num_hidden_layers)
else:
    self.d_hidden = d_hidden

neuralprophet/time_net.py
```diff
@@ -187,16 +187,14 @@ def __init__(
...
             for covar in self.config_covar.keys():
                 covar_net = nn.ModuleList()
-                d_inputs = self.n_lags
-                if self.config_covar[covar].as_scalar:
-                    d_inputs = 1
+                d_inputs = self.config_covar[covar].n_covars
                 for i in range(self.num_hidden_layers):
-                    covar_net.append(nn.Linear(d_inputs, self.d_hidden, bias=True))
-                    d_inputs = self.d_hidden
+                    d_hidden = d_inputs + n_forecasts if d_hidden is None else d_hidden
```
ourownstory, line 195: let's make this (separate from AR):
if d_hidden is None:
    self.d_hidden = max(4, (n_lags + n_forecasts) /(2.0* num_hidden_layers)
else:
    self.d_hidden = d_hidden

ourownstory (commented):
Looking good. 
I left some renaming and other detail suggestions.

neuralprophet/configure.py
ourownstory: if so, it may be better to name it n_lags :)

neuralprophet/df_utils.py
MateusGheorghe: Done

neuralprophet/forecaster.py
MateusGheorghe: Done

neuralprophet/time_net.py
MateusGheorghe: @ourownstory we have two problems with this suggestion: (1) we can have a 0 division when num_hidden_layers is not set (default = 0) and (2) we could have a max (4, float) , in other words, we could have, for instance, self.d_hidden = 4.5. 
I suggest this:
max(4, round((n_lags + n_forecasts)/(2.0*(num_hidden_layers+1)))

neuralprophet/time_net.py
MateusGheorghe: Same comment as above:
                    d_hidden = max(4, round((self.config_covar[covar].n_lags + n_forecasts)/(2.0*(num_hidden_layers+1)))) if d_hidden is None else d_hidden
replaced n_lags with self.config_covar[covar].n_lags so it does not depend on the AutoRegression lags

neuralprophet/forecaster.py
MateusGheorghe: I guess not. We only check for covars when we fit the model as we cannot add_lagged_regressors when instantiating NeuralProphet()

neuralprophet/forecaster.py
MateusGheorghe: Ok. However, it breaks the code in the assertion from time_dataset:
(line 264)                assert covar_config[covar].n_lags > 0

neuralprophet/forecaster.py
ourownstory: Yeah, makes sense. We should just make sure that this is not used before that function is called in fit.

PR #385, opened 2021-08-30
classification wrapper

MateusGheorghe:
Classification wrapper so NP can work with binary classification. The update depends on the previous pull request (n_regressors added to NP). The y column is the label of the problem and a lagged regressor is used as the input of the model (n_lags=0).


ourownstory:
carried on in #645

Issue #397, opened 2021-10-07
Allow Custom Loss function inheriting from _Loss


ourownstory:
### Discussed in https://github.com/ourownstory/neural_prophet/discussions/396

<div type='discussions-op-text'>

<sup>Originally posted by **nishai** October  7, 2021</sup>
Hi✌🏽
Firstly, thanks for all your work on `NeuralProphet`: it's a very cool tool!

# Problem
I'm struggling to train my `NeuralProphet` model with a custom loss function. I want to use the _Mean Pinball Loss_, and I've coded it up, replicating the [sklearn docs](https://scikit-learn.org/stable/modules/model_evaluation.html#pinball-loss), except using `torch` methods which will allow for autograd. However, I'm not able to use this custom loss function with `NeuralProphet`.

I've tested two different ways of passing the loss to `NeuralProphet`: (1) a simple python function, and (2) a custom class extending `torch.nn.modules.loss._Loss`, as advised [in this discussion](https://github.com/ourownstory/neural_prophet/discussions/212#discussioncomment-192052).

# Loss function: python function
```python
def MeanPinballLoss(y: Tensor, yhat: Tensor, alpha: int = 0.9) -> Tensor:
    y_diff = y - yhat
    yhat_diff = yhat - y

    loss = (
        alpha * torch.max(y_diff, torch.zeros_like(y_diff))
        + (1-alpha) * torch.max(yhat_diff, torch.zeros_like(yhat_diff))
    ).sum().mean()

    return loss
```

Using this function during model init `m = NeuralProphet(loss_func=MeanPinballLoss)`, I get the following error:
```
/opt/conda/lib/python3.7/site-packages/neuralprophet/configure.py in __post_init__(self)
...
NotImplementedError: Loss function <function MeanPinballLoss at 0x7f60b1656830> not found
```

I had a look at the source code, and from my understanding of `configure.py` 
https://github.com/ourownstory/neural_prophet/blob/31051a2cceb3aaefef3764f5e2a2181b14124f1f/neuralprophet/configure.py#L40-L55
this should not happen, since `callable(MeanPinballLoss)` is `True` (line 50).

# Loss function: PyTorch module loss

My next attempt was to extend `torch.nn.modules.loss._Loss`. I basically mimicked [PyTorch's implementation of L1Loss](https://github.com/pytorch/pytorch/blob/df11e2d6f9782fc3995e17ef09a5ef3812da041d/torch/nn/modules/loss.py#L30). 

```python
[...]

PR #406, opened 2021-10-11
Allow Custom Loss function inheriting from _Loss

nishai:
PR for Issue #397 

* Replaced check for existing `torch.nn.modules.loss` method, with a check for if the custom loss is a subclass of `torch.nn.modules.loss._Loss`.
* Added a test case to verify whether a custom loss extending `torch.nn.modules.loss._Loss` works.


ourownstory:
Looks good. will merge into a branch and run tests on my system, too.

PR #463, opened 2021-11-01
Df copy problem

MateusGheorghe:
The copy of lists of dataframes is performed in a more adequated way - [df_list.append(df_aux.copy(deep=True)) for df_aux in df]


neuralprophet/df_utils.py
```diff
@@ -150,7 +155,7 @@ def init_data_params(
     """
 
     if isinstance(df, list):
-        df_list = df.copy()
+        df_list = copy_list(df)
```
MateusGheorghe, line 158: Did not consider everyone as a list because of the log info for global normalization or local normalization

neuralprophet/df_utils.py
```diff
@@ -266,7 +271,7 @@ def normalize(df, data_params, local_modeling=False):
...
```
MateusGheorghe, line 274: Did not consider everyone as a list because of the merging and unmerging when considering global modeling

neuralprophet/df_utils.py
```diff
@@ -644,10 +651,8 @@ def fill_linear_then_rolling_avg(series, limit_linear, rolling):
...
 
 def make_list_dataframes(df, episodes):
-    df_list = list()
-    for i in range(0, episodes):
-        if df is not None:
-            df_list.append(df.copy())
-        else:
-            df_list.append(None)
+    if df is not None:
+        df_list = [df.copy(deep=True)] * episodes
```
ourownstory, line 655: why are we copying the dataframe here?

neuralprophet/df_utils.py
ourownstory: Which inputs are we not considering?

neuralprophet/df_utils.py
```diff
@@ -266,7 +271,7 @@ def normalize(df, data_params, local_modeling=False):
     """
 
     if isinstance(df, list):
-        df_list = df.copy()
+        df_list = copy_list(df)
         if local_modeling:
             # Local Normalization
             log.warning(
```
ourownstory, line 277: remove warning, assert len(data_params) == len(df_list)

MateusGheorghe (commented):
@ourownstory there you can find the changes you mentioned during our call.

PR #464, opened 2021-11-01
implmented csv logging functionality

bhausleitner:
two comments:

1. At the moment -> there is only one csvper experiment, so running experiments subsequently overwrites existing csv’s 
let’s discuss again the exact functionality again before I continue to implement dynamic storing, etc. 

2. `save_dir` and `save_metrics` are currently mandatory, shall we make them optional?


neuralprophet/benchmark.py
```diff
@@ -232,7 +234,25 @@ def __post_init__(self):
...
         }
 
-    def _evaluate_model(self, model, df_train, df_test):
+    def store_results_to_csv(self, fcst_train, fcst_test, save_dir, current_fold):
+        # save fcst and create dir if necessary
+        if save_dir:
+            folder_exist = os.path.isdir(save_dir)
+            if not folder_exist:
+                os.makedirs(save_dir)
+            filename_ext = (
```
ourownstory, line 243: why not use experiment name?

neuralprophet/benchmark.py
```diff
@@ -232,7 +234,25 @@ def __post_init__(self):
...
+                self.data.name
+                + "_"
+                + self.model_class.model_name
+                + "".join(["_{0}_{1}".format(k, v) for k, v in self.params.items()])
+                + ".csv"
+            )
+            if current_fold:
+                filename_ext = filename_ext[:-4] + "_fold_" + str(current_fold) + ".csv"
+            fcst_train.to_csv(self.save_dir + "/fcst_train_" + filename_ext, encoding="utf-8", index=False)
```
ourownstory, line 252: maybe good to use os.path.join()

neuralprophet/benchmark.py
```diff
@@ -232,7 +234,25 @@ def __post_init__(self):
             "SMAPE": _calc_smape,
         }
 
-    def _evaluate_model(self, model, df_train, df_test):
+    def store_results_to_csv(self, fcst_train, fcst_test, save_dir, current_fold):
+        # save fcst and create dir if necessary
+        if save_dir:
```
ourownstory, line 239: maybe better to be explicit: 
if save_dir is not None

neuralprophet/benchmark.py
```diff
@@ -277,7 +299,7 @@ def run(self):
         )
         model = self.model_class(self.params)
         model.fit(df=df_train, freq=self.data.freq)
-        result_train, result_test = self._evaluate_model(model, df_train, df_test)
+        result_train, result_test = self._evaluate_model(model, df_train, df_test, self.save_dir, None)
```
ourownstory, line 302: Why pass None to exp_name?

neuralprophet/benchmark.py
```diff
@@ -316,21 +339,24 @@ def run(self):
...
                 results_cv_test[m].append(result_test[m])
+            current_fold += 1
         return results_cv_train, results_cv_test
 
 
 @dataclass
 class Benchmark(ABC):
     """Abstract Benchmarking class"""
 
+    save_dir: str
```
ourownstory, line 359: let's make this optional

neuralprophet/benchmark.py
bhausleitner: experiment_name wraps the necessary data into a dict, where the key `params` is a dict converted to a  string. 
It's cleaner in my opinion, to directly grab the dict `params` and access the keys from there. Hmu if you don't know what I mean.

ourownstory: I agree. but we might want to also use this elsewhere. 
Let's add this string creation to __post_init__ 
Let's do this:
rename `experiment_name` to `experiment_metadata`
store this string you are creating as `experiment_name`

PR #507, opened 2021-12-22
Local normalization for global modeling

MateusGheorghe:


neuralprophet/forecaster.py
```diff
@@ -1223,7 +1283,13 @@ def _predict_raw(self, df, include_components=False):
                             component_vectors[name].append(value.detach().numpy())
 
         predicted = np.concatenate(predicted_vectors)
-        scale_y, shift_y = self.data_params["y"].scale, self.data_params["y"].shift
+        if self.local_modeling:
+            scale_y, shift_y = (
+                self.data_params.norm_params_dict[df_names]["y"].scale,
```
ourownstory, line 1288: beware - indexing with a list. this will likely not work.
We likely need to call predict_raw len(df_names) times and return it as a list/dict

neuralprophet/forecaster.py
```diff
@@ -1381,55 +1454,74 @@ def predict(self, df, decompose=True, raw=False):
...
         df = self._check_dataframe(df, check_y=False, exogenous=False)
-        df = df_utils.normalize(df, self.data_params, local_modeling=self.local_modeling)
+        df = df_utils.normalize(df, self.data_params, local_modeling=self.local_modeling, df_names=df_names)
         t = torch.from_numpy(np.expand_dims(df["t"].values, 1))
         trend = self.model.trend(t).squeeze().detach().numpy()
-        trend = trend * self.data_params["y"].scale + self.data_params["y"].shift
+        if self.local_modeling:
+            scale_y, shift_y = (
+                self.data_params.norm_params_dict[df_names]["y"].scale,
+                self.data_params.norm_params_dict[df_names]["y"].shift,
```
ourownstory, line 1480: see above about indexing with list

neuralprophet/forecaster.py
```diff
@@ -1381,55 +1454,74 @@ def predict(self, df, decompose=True, raw=False):
...
         """
-        df_list = df_utils.create_df_list(df)
+        df_list, df_names = df_utils.create_df_list(df)
+        if self.local_modeling:
+            df_utils.check_compatibility_df_names(self.df_names, df_names, df_list)
+            list_of_names = df_names
+        else:
+            if df_names is not None:
+                log.info("dict of DataFrames provided - Ignoring keys as local modeling is not being used")
+            list_of_names = [None] * len(df_list)
```
ourownstory, line 1503: use `else` to only create a None-list if df_names not None, otherwise keep names

neuralprophet/forecaster.py
ourownstory: or is it a type and it should not be a list but only a single name?

neuralprophet/forecaster.py
```diff
@@ -1381,55 +1454,74 @@ def predict(self, df, decompose=True, raw=False):
...
+        return df_forecast
 
-    def _predict_seasonal_components(self, df):
+    def _predict_seasonal_components(self, df, df_names=None):
         """Predict seasonality components
 
         Args:
             df (pd.DataFrame): containing column 'ds', prediction dates
-
+            df_names (list): list of names of dataframes provided (used for local modeling or local normalization)
```
ourownstory, line 1518: again, not a list?

neuralprophet/plot_model_parameters.py
```diff
@@ -26,7 +26,33 @@
     log.error("Importing matplotlib failed. Plotting will not work.")
 
 
-def plot_parameters(m, forecast_in_focus=None, weekly_start=0, yearly_start=0, figsize=None):
+def check_df_name(m, df_name):
```
ourownstory, line 29: what is this function used for?
isn't there a function that does the same but for a list - could you use that one for this or merge the two?

neuralprophet/plot_model_parameters.py
```diff
@@ -26,7 +26,33 @@
...
+            )
+        if df_name not in m.data_params.df_names:
+            raise ValueError(
+                "Global modeling local normalization was used. Name {name!r} missing from data params".format(
+                    name=df_name
+                )
+            )
+
+
+def plot_parameters(m, forecast_in_focus=None, weekly_start=0, yearly_start=0, figsize=None, df_name=None):
```
ourownstory, line 55: what happens if df_name is none? are the plots just in normalized scale?

tests/test_integration.py
```diff
@@ -818,8 +880,10 @@ def global_modeling_events_plus_regressors():  ### GLOBAL MODELLING + REGRESSORS
             for frst in forecast:
                 fig = m.PLOT(frst)
```
ourownstory, line 881: why capital letters?

tests/test_integration.py
```diff
@@ -729,6 +729,68 @@ def global_modeling():  ### GLOBAL MODELLING - NO EXOGENOUS VARIABLES
                     fig1 = m.plot(frst)
                     fig2 = m.plot_components(frst)
 
+    def global_modeling_local_normalization():  ### GLOBAL MODELLING - NO EXOGENOUS VARIABLES - LOCAL MODELING
+        m = NeuralProphet(n_forecasts=2, n_lags=10, epochs=EPOCHS, batch_size=BATCH_SIZE)
+        df_dict = {"dataset1": df1_0, "dataset2": df2_0}
+        df_dict2 = {"dataset1": df1_0, "dataset3": df3_0}
+        df_list = [df1_0, df2_0]
+        with pytest.raises(ValueError):
```
ourownstory, line 737: nice use of pytest.raises!

tests/test_integration.py
```diff
@@ -729,6 +729,68 @@ def global_modeling():  ### GLOBAL MODELLING - NO EXOGENOUS VARIABLES
                     fig1 = m.plot(frst)
                     fig2 = m.plot_components(frst)
 
+    def global_modeling_local_normalization():  ### GLOBAL MODELLING - NO EXOGENOUS VARIABLES - LOCAL MODELING
```
ourownstory, line 732: Could you separate the different global modelling test cases into separate test functions (not nested)?
This would make each one easier to understand.

ourownstory (changes requested):
Great work! I reviewed it in more detail than usual and added a few important and a few code-style suggestions.

neuralprophet/df_utils.py
```diff
@@ -265,31 +323,24 @@ def normalize(df, data_params, local_modeling=False):
...
-            df_joined, episodes = join_dataframes(df_list)
-            df = _normalization(df_joined, data_params)
-            df = recover_dataframes(df, episodes)
-    else:
+    df_list, _ = create_df_list(df)
+    if local_modeling:
+        # Local Normalization
+        df_list_norm = local_normalization(data_params, df_list, df_names=df_names)
+        df = df_list_norm[0] if len(df_list_norm) == 1 else df_list_norm
+    if not local_modeling and len(df_list) > 1:
```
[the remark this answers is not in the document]
MateusGheorghe: True! Way simpler.

neuralprophet/df_utils.py
```diff
@@ -843,3 +894,21 @@ def infer_frequency(df, freq, n_lags, min_freq_percentage=0.7):
...
+
+    Args:
+        former_df_names (list): usually the names provided to the dict of dataframes used in the fit procedure (self.df_names - train dict keys)
+        df_names (list): list of the names of dataframes provided for any of the procedures after the model is trained (keys of test dict)
+        df_list (pd.DataFrame or list of pd.DataFrame): data
+    """
+    if df_names is None:
+        raise ValueError("Please insert a dict with key values compatible with train dict")
+    else:
+        for name in df_names:
```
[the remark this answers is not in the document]
MateusGheorghe: Ok

neuralprophet/forecaster.py
```diff
@@ -4,6 +4,7 @@
 import pandas as pd
 
 import torch
+from torch._C import ErrorReport
```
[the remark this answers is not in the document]
MateusGheorghe: It might have passed after some updates of the master - deleted it.

neuralprophet/forecaster.py
```diff
@@ -493,14 +496,20 @@ def _init_train_loader(self, df):
                 regressor_config=self.regressors_config,
                 events_config=self.events_config,
                 local_modeling=self.local_modeling,
+                df_names=self.df_names,
             )
-        df = df_utils.normalize(df, self.data_params, local_modeling=self.local_modeling)
+        if self.local_modeling:
+            self.df_names = self.data_params.df_names
+        if not self.local_modeling and self.df_names is not None:
+            log.warning("Ignoring df_names as local normalization is not being used")
```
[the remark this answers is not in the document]
MateusGheorghe: ok

neuralprophet/forecaster.py
```diff
@@ -808,12 +827,18 @@ def split_df(self, df, freq="auto", valid_p=0.2, local_modeling=False):
...
+            local_modeling (bool): Defines the local normalization when using global modeling - each dataframe will be split according to valid_p locally
 
         Returns:
             df_train (pd.DataFrame):  training data
             df_val (pd.DataFrame): validation data
         """
-        df = df_utils.create_df_list(df)
+        if local_modeling and not isinstance(df, dict):
+            raise ValueError("Please insert a dict with dataframes in case of local normalization")
+        if isinstance(df, dict):
```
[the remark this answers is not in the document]
MateusGheorghe: I see, I am going to think about a way to receive and return dicts in the case that local_split is false.

neuralprophet/forecaster.py
```diff
@@ -825,6 +850,9 @@ def split_df(self, df, freq="auto", valid_p=0.2, local_modeling=False):
             inputs_overbleed=True,
             local_modeling=local_modeling,
         )
+        if local_modeling:
```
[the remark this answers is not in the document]
MateusGheorghe: Yes, the idea is if the user passes a dict we should return a dict. In case of a list, we return a list and so on. However, in case we receive a dict and do not do a local_split then it becomes a little bit more complicated. I will think about this.

neuralprophet/forecaster.py
```diff
@@ -843,7 +871,7 @@ def crossvalidation_split_df(self, df, freq="auto", k=5, fold_pct=0.1, fold_over
                 df_train (pd.DataFrame):  training data
                 df_val (pd.DataFrame): validation data
         """
-        if isinstance(df, list):
+        if isinstance(df, list) or isinstance(df, dict):
```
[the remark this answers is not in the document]
MateusGheorghe: Ok

neuralprophet/forecaster.py
```diff
@@ -962,16 +997,29 @@ def test(self, df):
...
         Returns:
             df with evaluation metrics
         """
+        df_list, df_names = df_utils.create_df_list(df)
+        if self.local_modeling:
+            df_utils.check_compatibility_df_names(self.df_names, df_names, df_list)
+            list_of_names = df_names
+        else:
+            if df_names is not None:
+                log.info("dict of DataFrames provided - Ignoring keys as local modeling is not being used")
```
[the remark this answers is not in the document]
MateusGheorghe: I forgot to update the doctrings, df_names were provided before (when dicts were not accepted).

neuralprophet/forecaster.py
```diff
@@ -1124,12 +1172,20 @@ def _prepare_dataframe_to_predict(self, df):
...
         return df
 
     def make_future_dataframe(self, df, events_df=None, regressors_df=None, periods=None, n_historic_predictions=False):
-        df_list = df_utils.create_df_list(df)
+        df_list, df_names = df_utils.create_df_list(df)
+        if self.local_modeling:
+            df_utils.check_compatibility_df_names(self.df_names, df_names, df_list)
+        else:
+            if df_names is not None:
+                log.info("dict of DataFrames provided - Ignoring keys as local modeling is not being used")
```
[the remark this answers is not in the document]
MateusGheorghe: Ok

neuralprophet/forecaster.py
MateusGheorghe: It is a typo, I changed for "df_name". _predict_raw is always called inside a loop in which we go over each dataframe name each time.

neuralprophet/forecaster.py
MateusGheorghe: Same as comment above

neuralprophet/forecaster.py
MateusGheorghe: df_names can be not None if a dict is passed but local_modeling is not being used. Therefore, a None-list is required. df_names is later used for the creation of the output dict.

neuralprophet/forecaster.py
MateusGheorghe: same as above

neuralprophet/plot_model_parameters.py
MateusGheorghe: I believe it is better if we have a different function in this case because we actually have to provide the df_name in this case. We do not extract it from the dict (as the forecaster functions cases) so the errors raised are different. Don't you agree?

neuralprophet/plot_model_parameters.py
MateusGheorghe: if df_name is none and local_modeling is true we raise an error (we can't plot the parameters unless we provide a name to refer to the data_params).

tests/test_integration.py
MateusGheorghe: I wonder the same haha I will change it.

tests/test_integration.py
MateusGheorghe: The only problem in this case is to load the peyton dataset over and over. I will separate them anyways. I will also add NROWS to the read_csv call so it makes the testing faster.

neuralprophet/forecaster.py
```diff
@@ -1124,12 +1172,20 @@ def _prepare_dataframe_to_predict(self, df):
             # fill in missing nans except for nans at end
             df = self.handle_missing_data(df, freq=self.data_freq, predicting=True)
         # normalize
-        df = df_utils.normalize(df, self.data_params, local_modeling=self.local_modeling)
+        if self.local_modeling:
```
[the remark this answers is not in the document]
MateusGheorghe: Done

neuralprophet/forecaster.py
```diff
@@ -493,14 +496,20 @@ def _init_train_loader(self, df):
...
                 events_config=self.events_config,
                 local_modeling=self.local_modeling,
+                df_names=self.df_names,
             )
-        df = df_utils.normalize(df, self.data_params, local_modeling=self.local_modeling)
+        if self.local_modeling:
+            self.df_names = self.data_params.df_names
+        if not self.local_modeling and self.df_names is not None:
+            log.warning("Ignoring df_names as local normalization is not being used")
+        df = df_utils.normalize(df, self.data_params, local_modeling=self.local_modeling, df_names=self.df_names)
```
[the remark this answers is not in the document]
MateusGheorghe: Done

neuralprophet/df_utils.py
```diff
@@ -14,17 +14,53 @@ class ShiftScale:
...
+        self.df_dict = dict(zip(self.df_names, self.df_list))
+        self.norm_params_dict = dict(zip(self.df_names, self.norm_params))
+
+
 def copy_list(df_list):
     df_list_copy = [df.copy(deep=True) for df in df_list]
     return df_list_copy
 
 
 def create_df_list(df):
```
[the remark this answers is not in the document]
MateusGheorghe: Yes, didn`t notice I could merge copy_list and create_df_list. Great point.

neuralprophet/df_utils.py
```diff
@@ -14,17 +14,53 @@ class ShiftScale:
...
+    """Creates a df_list with based on the df input. It converts a pd.DataFrame or dict to a list. It also stores dict keys in a list.
+    Args:
+        df (list, dict, or pd.DataFrame): containing df or group of dfs with column 'ds', 'y' with training data
+
+    Returns:
+        df_list: list of dataframes or copy of list of dataframes
```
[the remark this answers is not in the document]
MateusGheorghe: Now they get copied

neuralprophet/df_utils.py
```diff
@@ -14,17 +14,53 @@ class ShiftScale:
...
+        df_names: list of names of provided dataframes or None in case of list or pd.DataFrame as input
+    """
     if isinstance(df, list):
         df_list = copy_list(df)
-    else:
+        df_names = None
+    elif isinstance(df, dict):
```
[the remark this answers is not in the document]
MateusGheorghe: done

neuralprophet/plot_model_parameters.py
ourownstory: Sounds reasonable. If this is a sub-case of the other function, the other one could call this one? but if you think they should be separate that's ok, too.

neuralprophet/plot_model_parameters.py
ourownstory: I think that for plotting we could allow this - we would just have to print an info/warning statement that plots are in normalized scale.

tests/test_integration.py
ourownstory: Sounds good.

neuralprophet/plot_model_parameters.py
MateusGheorghe: Ok! Makes sense, would we still have the option to provide a name referring to data params?

neuralprophet/df_utils.py
```diff
@@ -147,7 +189,8 @@ def init_data_params(
             with sub_parameters normalize (bool)
         events_config (OrderedDict): user specified events configs
         local_modeling (bool): when set to true each episode from list of dataframes will be considered
-        locally (i.e. seasonality, data_params, normalization) - not fully implemented yet.
+        locally (i.e. seasonality, data_params, normalization)
```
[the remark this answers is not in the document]
MateusGheorghe: No, it is not a dict at this point. It can be a dict as input for the user, but it is converted to a list right after.

neuralprophet/plot_model_parameters.py
MateusGheorghe: @ourownstory I don't believe this is feasible. We can discuss this tomorrow.

MateusGheorghe: We can't predict the trend with no data_params related to df.

PR #541, opened 2022-02-12
Local normalization for panel datasets

ourownstory:
see #507


ourownstory:
@MateusGheorghe  I tried to restructure the use of dicts, but it is still in the middle - kind of a mess - let's discuss this in our next meeting.

PR #561, opened 2022-02-23
fix documentation parameter formatting

epistoteles:
Unfortunately, sphinx (or furo?) gets upset when you leave out a newline before `Args:` in the docstrings.

Because of this, some entries in the documentation looked like this:

param-formatting

To fix this, I have added newlines in front of all docstring `Args:` that were missing them. For consistency among all functions, I also added newlines before every `Returns:` (this is just a style choice and black does not care either).

I checked if this fixes the problem and it does:

param-formatting-after

I also corrected the formatting of parameter types in the `add_seasonality` function:

params

Please note that I have so far only checked and fixed the `forecaster.py` file. There may be more such mistakes/inconsistencies in other files.

There is also one big issue remaining in the `forecaster.py` file: the params of the `NeuralProphet` class itself are not shown in the documentation. This can be solved by moving up the docstring containing the param definitions to where the class begins. However, it also results in the comments (e.g. `## Trend Config`) being printed in the documentation. Therefore I decided not to include it in this PR.

This PR only contains the file  `forecaster.py`, not the generated html files.

### DIFF
--- a/neuralprophet/forecaster.py
+++ b/neuralprophet/forecaster.py
@@ -391,9 +391,9 @@ class NeuralProphet:
         Note: regularization and mode (additive/multiplicative) are set in the main init.
 
         Args:
-            name: string name of the seasonality component.
-            period: float number of days in one period.
-            fourier_order: int number of Fourier components to use.
+            name (string): name of the seasonality component.
+            period (float): number of days in one period.
+            fourier_order (int): number of Fourier components to use.
 
         Returns:
             The NeuralProphet object.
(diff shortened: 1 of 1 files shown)


ourownstory:
Thank you @epistoteles !
That makes quite the difference.

Please do make another PR for your suggested changes to the class docstring. 
We are trying to move towards generating the documentation mostly from docstrings, so your edits are very welcome.

ourownstory:
And if you manage to edit some of the other files' docstrings, that would also be wonderful.

PR #566, opened 2022-02-25
make class params visible and introduce sphinx comment command

epistoteles:
As mentioned in my last PR #561, the parameters of the `NeuralProphet` class were not visible in the documentation. I moved them up to the docstring directly under the class definition to make them visible.

In the docstring, some of the parameters had comments as headers, grouping them together, like the following one:
```
## AR Config
n_lags (int): Previous time series steps to include in auto-regression. Aka AR-order
ar_sparsity (float): [0-1], how much sparsity to enduce in the AR-coefficients.
    Should be around (# nonzero components) / (AR order), eg. 3/100 = 0.03
```

This double number sign is not recognized by sphinx as a comment, resulting in those lines being interpreted as parameters themselves. 

Unfortunately, sphinx (or rather: [reStructuredText](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html)s) own comment signifier `..` did also not work properly when I tried it out.

To preserve the comments (I think they are useful) I had to define a custom sphinx control sequence by following [this](https://stackoverflow.com/a/26537753/12416401) Stackoverflow answer. I adapted it to make sphinx ignore all text written between two occurrences of `COMMENT`:

```
COMMENT
AR Config
COMMENT
n_lags (int): Previous time series steps to include in auto-regression. Aka AR-order
ar_sparsity (float): [0-1], how much sparsity to enduce in the AR-coefficients.
    Should be around (# nonzero components) / (AR order), eg. 3/100 = 0.03
```

Unfortunately, the comment now occupies three lines instead of one - I found no better solution.

Like last time, this change only affects the `forecaster.py` file for now. I want to know if you consider this good style before making further changes.

I also want to mention that some parameter descriptions, like the following one, have highlighted parts of the text (because it is indented twice).

comment

I am not sure if the strong emphasis on those text segments was intended by you. I left them unchanged for now.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -25,7 +25,7 @@ from typing import Dict, Any
 # -- Project information -----------------------------------------------------
 
 project = "NeuralProphet"
-copyright = "2022, Oskar Triebe"
+copyright = "2021, Oskar Triebe"
 author = "Oskar Triebe"
 
 
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -15,7 +15,7 @@
 import os
 import sys
 import sphinx_fontawesome
-from sphinx.ext.autodoc import between
+from sphinx.ext.autodoc import between 
 
 # sys.path.insert(0, os.path.abspath('.'))
 sys.path.insert(0, os.path.abspath("../.."))
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -25,7 +25,7 @@ from typing import Dict, Any
 # -- Project information -----------------------------------------------------
 
 project = "NeuralProphet"
-copyright = "2021, Oskar Triebe"
+copyright = "2022, Oskar Triebe"
 author = "Oskar Triebe"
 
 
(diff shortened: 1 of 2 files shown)


ourownstory:
Hi @epistoteles Thank you for the fix and thoughtful comments!

Sphinx seems  a little moody here. Oh well. I think your workaround - works. 
Though the extra lines are not desirable, the advantages of the params showing up and us still being able to add extra comments are worth it.

Regarding emphasis - actually those sections were rather supposed to be de-emphasized. 
If you don't mind, it would be good if you could indent them to the same level such that they just show up as normal text.
I'll merge it after that.

Thank you!

epistoteles:
It is named sphinx because of the many riddles it gives you!

I unindented the mentioned sections.

I also noticed other strange behaviors. For example, the following return docstring

```
Returns:
    df_train (pd.DataFrame):  training data
    df_val (pd.DataFrame): validation data
```

gets rendered like this:

Tuple

You have to change it to the following:

```
Returns:
    tuple of two pd.DataFrames:
        df_train (pd.DataFrame):
            training data
        df_val (pd.DataFrame):
            validation data
```

to render it like this:

Tuple-after

So I also did that.

Based on how many strange formatting issues there are it could make sense to really sit down and go through one documentation page from top to bottom to identify them all.

PR #606, opened 2022-03-29
update requirement for dataclasses package

gkanapathy:
dataclasses is included in python 3.7+, so is not needed (or desired) there.


ourownstory:
Thank you @gkanapathy !

PR #619, opened 2022-04-17
Allow missing values left unimputed - drop affected areas

judussoari:
Missing values left unimputed are now allowed for the auto-regression flow. 
User gets to choose whether or not they want NaN values/windows to be dropped from the data.

Now, impute_missing (int, bool) describes the maximum number of values to be auto-imputed. Default: 20 
In case of (consecutive) NaN windows larger than impute_missing: inputs and targets with NaN may be dropped from the data. Default: no dropping. 

Possible consequences of missing data are: 
-Distorted/poor training/validation results due to a lot of dropped NaN values in the dataframes
-Prediction fails, if future_dataframe contains NaN values. ValueError will be raised in this case


ourownstory:
Great work!
Based on our discussion, final adjustments
- first run linear imputation (skip rolling avg) (use impute_missing as int arg)
- drop samples with missing targets
- log.warning if samples are dropped (tell how many missing, how many dropped)
- write docs in docstrings
- skip ratio calculation (#370), do in separte PR later

ourownstory:
updates
- support impute_missing as boolean
- move dropping of nan values to init_after_tabularized TimeDataset, with new arg: drop_nan_samples

yanirs:
Sorry for the drive-by comment, but I implemented a workaround for the same issue locally and thought it might be useful.

The workaround is very simple &ndash; it exposes the `impute_limit_linear` and `impute_rolling` in an extended subclass that I use instead of `NeuralProphet`:

```python
class ExtendedNeuralProphet(neuralprophet.NeuralProphet):
    def __init__(
        self,
        impute_limit_linear: int = 5,
        impute_rolling: int = 20,
        **kwargs: Any
    ):
        super().__init__(**kwargs)
        self.impute_limit_linear = impute_limit_linear
        self.impute_rolling = impute_rolling
```

I understand that this PR goes further by supporting training without imputing the missing values, but it seems like this may lead to surprising results once this is merged, as NaNs can now be dropped.

I wonder if it'd be better to make the behaviour more explicit and backwards-compatible with a new setting of this type:

```python
@dataclass
class MissingDataHandling:
    impute_missing: bool = True
    impute_limit_linear: int = 5
    impute_rolling: int = 20
    drop_nan_samples: bool = False
```

Explanation:

* `missing_data_handling: MissingDataHandling` would be accepted by `NeuralProphet.__init__()`.
* With the default settings, the behaviour would be the same as it is right now: Missing values are imputed with the `impute_limit_linear` and `impute_rolling` arguments, and an exception is raised if any missing values remain.
* `impute_limit_linear` and `impute_rolling` can be changed to adjust the imputation settings, as in my workaround.
* NaN samples are dropped after imputation only if `drop_nan_samples` is True.

Thoughts? Sorry again if this has already be discussed. :slightly_smiling_face:

ourownstory:
> I implemented a workaround for the same issue locally and thought it might be useful.

Hi @yanirs 

Thank you for contributing your solution and your thoughtful suggestions!

@julioare94 
Let us actually use the dataclass Yanir is suggesting, adding it to `configure.py`.
However, we should not make it an expected input to the NP model init - just have the kwargs be forwarded to this class. This is more consistent with the other args.
(@yanirs we might actually build a modular 'advanced' API that directly accepts the config dataclasses, and allows for arbitrary module configuration, following some abstraction standard, at some point in the future)

Further, let us keep the `impute_rolling` procedure like Yanir is suggesting, for backwards compatibility. Currently It is not implemented in your PR afaik. 

Finally, also following Yanir's suggestion, we can instead of activating the dropping of samples by default, raise an error message that points the user to this setting, if they have too many missing values. 
We can later activate this by default once we also have the 'skip ratio' calculation (#370) implemented.

ourownstory:
@yanirs merged. thank you for your suggestions!

PR #628, opened 2022-04-27
update future regressors doc

bhausleitner:


neuralprophet/forecaster.py
```diff
@@ -433,6 +433,11 @@ def add_future_regressor(self, name, regularization=None, normalize="auto", mode
             mode : str
                 ``additive`` (default) or ``multiplicative``.
 
+                Note
```
yanirs, line 436: It's probably worth adding this as a top-level note rather than under `mode`.

Also, I think you meant i.e. (_that is_) rather than e.g. (_for example_) &ndash; see [this handy explanation](https://www.merriam-webster.com/words-at-play/ie-vs-eg-abbreviation-meaning-usage-difference). :slightly_smiling_face:

yanirs (commented):
Thanks @bhausleitner! I think it's also worth adding the same note to the top of https://raw.githubusercontent.com/ourownstory/neural_prophet/main/docs/source/future-regressors.md.

PR #642, opened 2022-05-19
add example for crossvalidation

MateusGheorghe:


neuralprophet/df_utils.py
```diff
@@ -639,7 +639,7 @@ def _crossvalidation_with_time_threshold(df_dict, n_lags, n_forecasts, k, fold_p
     return folds
 
 
-def crossvalidation_split_df(df, n_lags, n_forecasts, k, fold_pct, fold_overlap_pct=0.0, global_model_cv_type="None"):
+def crossvalidation_split_df(df, n_lags, n_forecasts, k, fold_pct, fold_overlap_pct=0.0, global_model_cv_type=None):
```
ourownstory, line 642: What was the reason for `global_model_cv_type=None` instead of `global_model_cv_type="global-time"`?

MateusGheorghe: Done @ourownstory

Issue #498, opened 2021-12-13
Auto-Regression fails when there is NaN in recent data


winedarksea:
```ValueError: need at least one array to concatenate```

Issue only occurs when there is NaN - I believe NaN data is meant to be handled by default, but if not, then you can ignore this Issue. 

### Code to Reproduce:
```python
import numpy as np
import pandas as pd
from neuralprophet import NeuralProphet
def load_sine(shape, start_date: str = "2021-01-01"):
    df_wide = pd.DataFrame(
        np.ones(shape),
        index=pd.date_range(start_date, periods=shape[0], freq="D"),
        columns=range(shape[1]),
    )
    X = pd.to_numeric(df_wide.index, errors='coerce', downcast='integer').values

    def sin_func(a, X):
        return a * np.sin(1 * X) + a

    for column in df_wide.columns:
        df_wide[column] = sin_func(column + 1, X)
    return df_wide

df = load_sine(shape=(300, 1)).reset_index(drop=False)
df.columns = ['ds', 'y']
df.iloc[-1, 1] = np.nan
forecast_length = 10

m = NeuralProphet(
    n_lags=forecast_length * 2,
    n_forecasts=forecast_length,
)

metrics = m.fit(df, freq='D')

df_future = m.make_future_dataframe(df, periods=forecast_length)
forecast = m.predict(df_future, decompose=False)
```

comment out this line and it works: ```df.iloc[-1, 1] = np.nan```

winedarksea:
On a related note, `n_forecasts` only matters when `n_lags` is set, right?

ourownstory:
Thank you @winedarksea for filing this bug, and sorry for not responding sooner.

I ran your code, and it seems you found a special edge case that produces a bug.
- Only occurs if `NaN` is at very end (not in middle or very beginning)
- Only if `make_future_dataframe` is called before, and
- Only if `make_future_dataframe` is called with `periods > 0` and `n_historic_predictions == 0 or False`

We will have a closer look at this.

Thank you,
Okar

ourownstory:
Hi @winedarksea Yeah, I agree, it's a rather common edge case. We will update our tests to include this when we dive into a solution. 

Btw, would you mind sharing how you would expect missing values at the end to be filled in?

Options that come to mind:
minimal fix:
* last observed value repeated for all missing values

hard to generalize:
* average over last N, e.g. `n_lags` values (challenge: 'reasonable' length of past to take into account varies)
[...]

PR #653, opened 2022-06-06
Fixed Issue #498

judussoari:
Any NaN value in df_future will be forward-filled. 
Despite being a very simple solution, this avoids the error message that has been reported in Issue #498. Additionally, this fix ensures we never run m.predict() with NaN values in df_future, which previously made the predictions useless.

### DIFF
--- a/neuralprophet/forecaster.py
+++ b/neuralprophet/forecaster.py
@@ -2266,13 +2266,15 @@ class NeuralProphet:
             df = pd.DataFrame(columns=df.columns)
         else:
             df = df[-(self.max_lags + n_historic_predictions) :]
+            # FIX ISSUE 498: forward-fill any NaN values in df_future
             if np.isnan(df["y"]).any():
-                raise ValueError(
-                    "Data used for historic forecasts contains NaN values. "
-                    "Please ensure there are no NaN values within the last {} entries of the df".format(
-                        self.max_lags + n_historic_predictions
-                    )
-                )
+                df = df.fillna(method="ffill")
+                # raise ValueError(
+                #     "Data used for historic forecasts contains NaN values. "
+                #     "Please ensure there are no NaN values within the last {} entries of the df".format(
+                #         self.max_lags + n_historic_predictions
+                #     )
+                # )
 
         if len(df) > 0:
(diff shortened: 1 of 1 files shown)


neuralprophet/forecaster.py
```diff
@@ -2265,14 +2265,20 @@ def _make_future_dataframe(self, df, events_df, regressors_df, periods, n_histor
         if (n_historic_predictions + self.max_lags) == 0:
             df = pd.DataFrame(columns=df.columns)
         else:
+            # FIX ISSUE #498
+            last_valid_index = df["y"].last_valid_index()
+            last_index = df["y"].index[-1]
+            last_non_nan_index = last_index - last_valid_index + 1
```
ourownstory, line 2271: Doesn't `df["y"].last_valid_index()` already return this?

neuralprophet/forecaster.py
```diff
@@ -2265,14 +2265,20 @@ def _make_future_dataframe(self, df, events_df, regressors_df, periods, n_histor
...
-                raise ValueError(
-                    "Data used for historic forecasts contains NaN values. "
-                    "Please ensure there are no NaN values within the last {} entries of the df".format(
-                        self.max_lags + n_historic_predictions
-                    )
-                )
+            nan_at_end = 0
+            while len(df) > nan_at_end and df["y"].isnull().iloc[-(1 + nan_at_end)]:
+                nan_at_end += 1
+            fill_thresh = nan_at_end - n_historic_predictions
```
ourownstory, line 2276: Shouldn't this be connected to n_lags instead of n_historic_predictions?

neuralprophet/forecaster.py
```diff
@@ -2265,14 +2265,20 @@ def _make_future_dataframe(self, df, events_df, regressors_df, periods, n_histor
...
+            if fill_thresh > 0:
```
ourownstory, line 2277: Why only if there are more nans than historic prections?
Should we not always do this? (except if there is some excessive amount of missing values - and then we should trow a warning or error)

ourownstory (changes requested):

ourownstory (approved):

Issue #189, opened 2020-12-08
Data: Positivity constraint for data preprocessing - replace/remove negative samples if activated


ourownstory:
see discussion in #141

PR #658, opened 2022-06-16
Future regressor positivity constraint - Issue #189

karl-richter:
**Status quo:**
Future regressors are currently not constrained or sanity checked to fulfil any criteria.
However, this functionality is requested by #189.

This pr implements the data positivity constraint that was proposed by @ourownstory in the discussion:
"Nevertheless, this could be avoided by throwing a warning/error if a user sets a positivity constraint and then provides negative training values." [Discussion](https://github.com/ourownstory/neural_prophet/discussions/141#discussioncomment-169682)

**Changes:**
- Extended the `add_future_regressor` function with the attribute `constraint` which can take the values `None (default)`  or `positive`
- Extended the Regressor class with a `constraint` attribute
- Extended the `__handle_missing_data` function to check for negative values and raise a `ValueError` in case.

**Open questions:**
- Is raising a `ValueError` the expected behaviour? I would argue yes, as with this implementation the user would explicitly state that he expects only positive values to be in the df and therefore should be informed.
- Is `__handle_missing_data` the appropriate place to check the constraint? It seems to be the most straight forward as the regressors are anyways checked in the function but one could argue that it is not really related to missing data.
- Should further constraints be implemented (eg. a negative-constraint or non-zero)?
- Should this behaviour be covered in any of the test cases (eg. `test_future_reg`)?

### DIFF
--- a/neuralprophet/forecaster.py
+++ b/neuralprophet/forecaster.py
@@ -442,7 +442,7 @@ class NeuralProphet:
             )
         return self
 
-    def add_future_regressor(self, name, regularization=None, normalize="auto", mode="additive"):
+    def add_future_regressor(self, name, regularization=None, normalize="auto", mode="additive", constraint=None):
         """Add a regressor as lagged covariate with order 1 (scalar) or as known in advance (also scalar).
         The dataframe passed to :meth:`fit`  and :meth:`predict` will have a column with the specified name to be used as
         a regressor. When normalize=True, the regressor will be normalized unless it is binary.
@@ -465,6 +465,8 @@ class NeuralProphet:
                 if ``auto``, binary regressors will not be normalized.
             mode : str
                 ``additive`` (default) or ``multiplicative``.
+            constraint : str
+                optional, allows to add a constraint to the regressor. ``None`` (default) or ``positive``.
 
 
         """
@@ -479,7 +481,9 @@ class NeuralProphet:
 
         if self.regressors_config is None:
             self.regressors_config = {}
-        self.regressors_config[name] = configure.Regressor(reg_lambda=regularization, normalize=normalize, mode=mode)
+        self.regressors_config[name] = configure.Regressor(
+            reg_lambda=regularization, normalize=normalize, mode=mode, constraint=constraint
+        )
         return self
 
     def add_events(self, events, lower_window=0, upper_window=0, regularization=None, mode="additive"):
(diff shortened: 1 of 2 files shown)


karl-richter:
The codecoverage seems to have been decreased as the if-statement is currently not covered by a pytest. Should this be adapted by eg. introducing a new test that checks for the ValueError to arise in case a df with negative samples is passed?

ourownstory:
> The codecoverage seems to have been decreased as the if-statement is currently not covered by a pytest. Should this be adapted by eg. introducing a new test that checks for the ValueError to arise in case a df with negative samples is passed?

@karl-richter on second thought: As the Error is actually intended behavior, it may make sense to add such a check. You can use the `with pytest.raises(ValueError):` syntax.

ourownstory:
> * Extended the `add_future_regressor` function with the attribute `constraint` which can take the values `None (default)`  or `positive`

Great. Let's also add this to `lagged_regressor`, Oh, and please add this to the series itself (y) in a separate PR. 
Ideally all use the same function to do the checks/replacements, that you call in handle missing multiple times.

Thinking about this, the constraint should be more for outputs - as we are just dealing with inputs in this PR, lets:
* rename `constraint` to `fill_negative`
* assert that the   `fill_negative` is one of the permitted values
* permitted options: None, float, 'missing', 'error'

> * Extended the Regressor class with a `constraint` attribute

The assertion can happen in the `post_init` of this class

> **Open questions:**
> * Is raising a `ValueError` the expected behaviour? I would argue yes, as with this implementation the user would explicitly state that he expects only positive values to be in the df and therefore should be informed. Alternatively, negative values could be removed / replaced with zero.

I am not so sure about this. let's go with the permitted options above. that will allow to do nothing, fill with a value, remove them (make them missing values), or throw an error. I think missing or fill could be most useful.


> * Is `__handle_missing_data` the appropriate place to check the constraint? It seems to be the most straight forward as the regressors are anyways checked in the function but one could argue that the feature is not really related to missing data.

Yes.

> * Should further constraints be implemented (eg. a negative-constraint or non-zero)?

Answered above (permitted options)

> * Should this behaviour be covered in any of the test cases (eg. `test_future_reg`)?

Yes, it would be great if you could add a new test function.

karl-richter:
Hi @ourownstory, I addressed the proposed changes and adapted the implementation.

- renamed the parameter from `constraint` to `handle_negatives` (to me this explains better what the parameter does than `fill_negative`. Do you agree? Otherwise I can switch it back to `fill_negative`)
- extended the implementation to support filling, removal and error raising
- added an integration test to cover the negative filling and removal behaviour
- added a unit test to cover the error raising behaviour in case of `handle_negatives=error`

neuralprophet/forecaster.py
```diff
@@ -406,6 +400,10 @@ def add_lagged_regressor(
             normalize : bool
                 optional, specify whether this regressor will benormalized prior to fitting.
                 if ``auto``, binary regressors will not be normalized.
+            handle_negatives : str, float, int
```
ourownstory, line 403: It may make sense to structure this docstring as a list of options that shows up nicely in the docs, e.g.         [trend_reg_threshold : bool, optional](https://github.com/ourownstory/neural_prophet/blob/93b26a04c640a2589af634f115d2e6c0204e6d8e/neuralprophet/forecaster.py#L79)

ourownstory (changes requested):

ourownstory:
@karl-richter Do you mind also adding this functionality to the series itself ('y')?

karl-richter:
@ourownstory the changes are addressed, I moved the function into `df_utils.py` and added the preprocessing functionality to the `fit()` method for `y`.

karl-richter:
Reversed the deep integration with NP functions (such as `add_lagged_regressor`, `add_future_regressor` and `fit`) based on discussions with @ourownstory. Instead, provide a shallow helper function to pre-process and constrain input data.

Issue #318, opened 2021-04-25
Create predict_last_forecast equivalent to plot_last_forecast (Get results as csv after forecasting with n_lags)


artem1480:
How to get result? y cloumn and yhat are NAN
![image](https://user-images.githubusercontent.com/60170640/116005492-dac50f80-a60f-11eb-9081-229a2baccb08.png)

artem1480:
![image](https://user-images.githubusercontent.com/60170640/116005544-1cee5100-a610-11eb-9882-6b1be2b3f588.png)

artem1480:
![IMG_20210425_223933.jpg](https://user-images.githubusercontent.com/60170640/116007184-24fdbf00-a617-11eb-8af8-40f6b051e03e.jpg)

Shiivong27:
If you're using n_forecasts = 10 and n_lags = 10 (for e.g.), then when you are making your future dataframe with the 'make_future_dataframe' method, be sure to define 'df' as your training set, 'regresors_df' as your test set, 'periods' as the length of the test set and set 'n_historic_predictions' to true.

When you forecast on the test dataset, you'll end up with predictions ranging from yhat1 to yhat10 (since n_forecasts is 10) with most of them being None. Plot the yhat10 values as they'll be the latest predictions of the model.

Hope that helps.

ourownstory:
Thank you @Shiivong27 

@artem1480 might you be looking for a version of plot_last_forecast that returns the df with the last predictions instead?

artem1480:
> Thank you @Shiivong27 
> 
> @artem1480 might you be looking for a version of plot_last_forecast that returns the df with the last predictions instead?

Perhaps. Where i can get it?

ourownstory:
> > Thank you @Shiivong27
> > @artem1480 might you be looking for a version of plot_last_forecast that returns the df with the last predictions instead?
> 
> Perhaps. Where i can get it?
It currently does not exist. Would you like to solve this issue by contributing it? 
You could use the plot_last_forecast as base code and just remove the plotting and return the forecast instead.

artem1480:
> > > Thank you @Shiivong27
> > > @artem1480 might you be looking for a version of plot_last_forecast that returns the df with the last predictions instead?
> > 
> > Perhaps. Where i can get it?
> 
> It currently does not exist. Would you like to solve this issue by contributing it?  :)
> You could use the plot_last_forecast as base code and just remove the plotting and return the forecast instead.
> 

[...]

PR #677, opened 2022-07-04
Add predict_last_forecast() to forecaster.py

FubonChu:
Add predict_last_forecast() to forecaster.py based on plot_last_forecast(), but returns a df.


judussoari:
I think this PR looks good @FubonChu ! I was actually working on the same issue without realizing you have already published a PR on it (sorry!)

Regarding the failing codecov/patch test: I am not sure why it is failing even though you are testing all of the possible cases that can occur in `get_latest_forecast`. 

In my draft PR #700 I reached 100% of diff hit with the same code. Only difference is that I wrote `assert len(fcst["ID"].unique()) > 1` instead of `assert fcst["ID"].unique() > 1`. I even used your integration test because I thought it was well-written (again sorry). 

Looping in @ourownstory to:
- confirm whether it's ok to return the df with the last predictions, or if a .csv shall be exported instead, as the title of Issue #318  suggests? 
- Maybe come up with another suggestion to pass the codecov/patch test

FubonChu:
Hi @judussoari , thank you for working on it. I have been struggled by it for long time. With your comments and assurance, I gave a try again by creating a new clean branch for the issue and it works now! New pull request is named #701 with the same codes for get_latest_forecast function, test and short tutorial.

By the way, as the `assert fcst["ID"].unique() > 1` has been used throughout the test_integration, I didn't change it for consistence. I am closing this PR now.

PR #656, opened 2022-06-12
Adressed pytest warnings - Issue #599

karl-richter:
Requested by #599.

**Status quo**
When running `pytest -v` on main, the pytest summary currently yields 626 warnings.

**Changes**
This pr resolves all warnings by refactoring the code according to the pandas docs:
- 615 Warnings: Replaced frame.append with pd.concat, according to [Pandas Deprecation Note 1.4.0](https://pandas.pydata.org/pandas-docs/dev/whatsnew/v1.4.0.html#deprecated-dataframe-append-and-series-append)
- 8 Warnings: Replaced dataframe slicing (eg. df["abc"] = ...) with pd.concat, according to [Appending rows to a DataFrame](https://pandas.pydata.org/pandas-docs/stable/user_guide/merging.html#appending-rows-to-a-dataframe)
- 1 Warning: Replaced dataframe slicing for value assignment with frame.loc, according to [Setting values in a DataFrame](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.loc.html)
- 2 Warnings: Replace the `interpolation=` arg of np.quantile with `method=`, according to [Numpy Deprecation Note 1.22.0](https://numpy.org/devdocs/release/1.22.0-notes.html#add-new-methods-for-quantile-and-percentile)
- 7 Warnings: Replaced dataframe slicing (eg. df["abc"]="xyz") with pd.assign (eg. df = df.assign(abc="xyz")), according to [Assigning columns in dataframe](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.assign.html)

PR #696, opened 2022-07-24
FIX: NaN IDs output in the predict method for Multiple Time Series

alfonsogarciadecorral:
Hi @ourownstory @karl-richter ,

When using the global_modeling tutorial I realised that when forecasting multiple time series, the output of the `predict` method gives some NaNs IDs (see picture attached).

I had a look and the problem might come from #656 and, in particular, from the method `_reshape_raw_predictions_to_forecst_df`. At the end of this method we have:

`df_forecast = pd.concat([df_forecast, pd.Series(yhat, name=comp)], axis=1, ignore_index=False)`

the problem is that the index in `df_forecast` and `pd.Series(yhat, name=comp)` do match if only 1 time series is passed or in the first time series passed when passing multiple. However the `df_forecast` of the second time series passed, has an indexation that wont match the `pd.Series`.

I think resetting the index at the start of the method should work. Let me know what you think!

WhatsApp Image 2022-07-24 at 6 45 43 PM

### DIFF
--- a/neuralprophet/forecaster.py
+++ b/neuralprophet/forecaster.py
@@ -2599,7 +2599,6 @@ class NeuralProphet:
         # Receives df with single ID column
         assert len(df["ID"].unique()) == 1
         cols = ["ds", "y", "ID"]  # cols to keep from df
-        df = df.reset_index(drop=True)
         df_forecast = pd.concat((df[cols],), axis=1)
         # create a line for each forecast_lag
         # 'yhat<i>' is the forecast for 'y' at 'ds' from i steps ago.
@@ -2635,5 +2634,7 @@ class NeuralProphet:
                 forecast_0 = components[comp][0, :]
                 forecast_rest = components[comp][1:, self.n_forecasts - 1]
                 yhat = np.concatenate(([None] * self.max_lags, forecast_0, forecast_rest))
-                df_forecast = pd.concat([df_forecast, pd.Series(yhat, name=comp)], axis=1, ignore_index=False)
+                df_forecast = pd.concat(
+                    [df_forecast, pd.Series(yhat, name=comp).set_axis(df_forecast.index)], axis=1, ignore_index=False
+                )
         return df_forecast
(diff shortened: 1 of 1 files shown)


ourownstory:
Great catch @alfonsogarciadecorral !
Yeah, that sounds like a good solution - did it fix the issue you observed?

Actually, in this case it may be the safer option to add a `ignore_index=True` to `concat`. 
Because we want the index to be ignored. Else it merges based on the index (which in our case has no meaning at best and at worst, like you found, leads to issues).
https://stackoverflow.com/questions/32801806/pandas-concat-ignore-index-doesnt-work

alfonsogarciadecorral:
Hey @ourownstory  I tried that, but column names were lost.

[https://stackoverflow.com/questions/71805129/pandas-dataframe-concatenation-with-axis-1-lost-column-names](https://stackoverflow.com/questions/71805129/pandas-dataframe-concatenation-with-axis-1-lost-column-names)

Screenshot from 2022-07-26 10-26-32

ourownstory:
Oh, why are we concatenating a tuple of one here?
And why along axis 1?
@karl-richter what was the original code here?
Something seems off

karl-richter:
Hi @alfonsogarciadecorral, thanks for the great catch!
This one is actually a bit tricky.

So @ourownstory, the original code there was actually super straight forward, we simply add a column to the data frame. The exact code was `df_forecast[comp] = yhat`. 

As mentioned in the stack overflow thread you shared, the ignore_index in pd.concat would simply remove the column names as tell it to concatenate along axis 1 (axis=0 would mean we add a row to the df), but does not allow to ignore the indices of the rows (what we would like it to do).

As @alfonsogarciadecorral already mentioned, resetting the index of the df prior to the concatenation solves the issue but does not preserve the index of the df. A possible solution would be to set the index of the series so that it matches the index of the df during the concatenation. This can be archived using `pd.Series.set_axis(df.index)`. I shared a small example that hopefully makes our discussion a bit more tangible.

https://gist.github.com/karl-richter/95296c6123685adb725baefde1bca245

Would we be happy with the fix proposed in the notebook? I would then check if any other parts of the code are affected.

alfonsogarciadecorral:
Hey @karl-richter 

Sorry for the late response. 

I agree! It should not be different to reset index for this particular case, but I see keeping the original indexing as a better practice.

ourownstory:
Great work guys!

Alfonso, as discussed, let's split it into two lines and add your original solution as comment to retain the reasoning in the code.

PR #705, opened 2022-08-05
[FIX] `infer_frequency` for monthly, yearly and quarterly data

### LINKED ISSUE #692: KeyError when creating a global model using freq='MS' for fit function
**Prerequisites**

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in a new virtualenv with only neuralprophet installed, directly from github:
      ```shell
      git clone <copied link from github>
      cd neural_prophet
      pip install .
      ```
    * Checked the Answered Questions on the Github Disscussion board: https://github.com/ourownstory/neural_prophet/discussions
      If you have the same question but the Answer does not solve your issue, please continue the conversation there.
    * Checked that your issue isn't already filed: https://github.com/ourownstory/neural_prophet/issues
      If you have the same issue but there is a twist to your situation, please add an explanation there.
    * Considered whether your bug might actually be solveable by getting a question answered: 
      * Please [post a package use question](https://github.com/ourownstory/neural_prophet/discussions/categories/q-a-get-help-using-neuralprophet)
      * Please [post a forecasting best practice question](https://github.com/ourownstory/neural_prophet/discussions/categories/q-a-forecasting-best-practices)
      * Please [post an idea or feedback](https://github.com/ourownstory/neural_prophet/discussions/categories/ideas-feedback)

**Describe the bug**

Bug Description:
When I run the fit function and use freq='MS' as a parameter I am getting a "KeyError: 0" and I suspect something is wrong with line 1195 in the _infer_frequency function in the df_utils.py file

**To Reproduce**

1. With the data 'the data from the energy_data_example.ipynb file'
2. Setting the model hyperparameters 'use defaults'
3. After running these code lines 'first change the df_ercot data such that it only includes the first day of each month and get rid of the hour stamp to prep data for the fit function. Also, remove rows where there are duplicates (make sure there is only one row for each first day of month date)'
4. When using this function 'run all the lines and change the fit function parameter 'freq' to be 'MS' instead of 'H''
5. See error

**Expected behavior**

I expected the global model to run as normal and fit my data with no errors.
**What actually happens**

Describe what happens, and how often it happens.
I get a Key Error: 0 and It happens everytime I use freq = "MS" in the fit function

**Screenshots**
Error Messages screenshots:
image
From forecaster.py:
image
From df_utils.py
image
image
image
Where I think the bug is:
image

If applicable, add screenshots and console printouts to help explain your problem.

**Environement (please complete the following information):**

 - Python environment [e.g. Python 3.8, in standalone venv with no other packages]
 - NeuralProphet version and install method [e.g. 2.7, installed from PYPI with `pip install neuralprophet`]
I downloaded installed the package in a standalone venv directly from this GitHub
**Additional context**

FedericoGarza:
Fix #692.

The problem in #692 arises because, for each series, the code tries to obtain the first datestamp using `df['ds'][0]`. However, when there is a data frame in a long format (with ID), potentially not all series share the same index (not all start at `0`). Therefore, it is best to use `df['ds'].iloc[0]`.

### DIFF
--- a/neuralprophet/df_utils.py
+++ b/neuralprophet/df_utils.py
@@ -1192,12 +1192,12 @@ def _infer_frequency(df, freq, min_freq_percentage=0.7):
     if frequencies[np.argmax(distribution)] == 2.6784e15 or frequencies[np.argmax(distribution)] == 2.592e15:
         dominant_freq_percentage = get_dist_considering_two_freqs(distribution) / len(df["ds"])
         num_freq = 2.6784e15
-        inferred_freq = "MS" if pd.to_datetime(df["ds"][0]).day < 15 else "M"
+        inferred_freq = "MS" if pd.to_datetime(df["ds"].iloc[0]).day < 15 else "M"
     # exception - yearly df (365 days freq or 366 days freq)
     elif frequencies[np.argmax(distribution)] == 3.1536e16 or frequencies[np.argmax(distribution)] == 3.16224e16:
         dominant_freq_percentage = get_dist_considering_two_freqs(distribution) / len(df["ds"])
         num_freq = 3.1536e16
-        inferred_freq = "YS" if pd.to_datetime(df["ds"][0]).day < 15 else "Y"
+        inferred_freq = "YS" if pd.to_datetime(df["ds"].iloc[0]).day < 15 else "Y"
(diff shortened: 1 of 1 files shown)


ourownstory:
Thank you @FedericoGarza for resolving this bug!

Issue #133, opened 2020-11-30
events/holidays regularization does not reduce weights to zero.


ourownstory:

noxan:
Happy to help contribute to this issue as discussed.

I've traced down the implementation to these functions:
1) https://github.com/ourownstory/neural_prophet/blob/bae43ab9f934eee81d79fb773ae2852604b81a54/neuralprophet/forecaster.py#L2025-L2028
2) https://github.com/ourownstory/neural_prophet/blob/bae43ab9f934eee81d79fb773ae2852604b81a54/neuralprophet/utils.py#L58

Documentation is currently lacking in NeuralProphet, the [documentation from Prophet](https://facebook.github.io/prophet/docs/seasonality,_holiday_effects,_and_regressors.html#prior-scale-for-holidays-and-seasonality) seems helpful.

To experiment with the effect of regularization on events/holidays, I've created the following scripts, including first results attached:
1) [Event regularization sample](https://github.com/noxan/neural_prophet/blob/issue-133/untitled-01.py), here without regularization the event weight is `OrderedDict([('birthday_+0', tensor(0.2050, grad_fn=<SelectBackward0>))])`, while for `regularization=1` it is reduced to `OrderedDict([('birthday_+0', tensor(6.3685e-07, grad_fn=<SelectBackward0>))])`
2) [Holiday regularization sample](https://github.com/noxan/neural_prophet/blob/issue-133/Untitled-1.ipynb), including [results for different regularization parameters](https://github.com/noxan/neural_prophet/blob/issue-133/regularization-results.txt)

PR #707, opened 2022-08-08
Write test cases for events and holidays regularization

noxan:
A collection of test cases to check regularization functions. Contains some unit tests for base functions and several integration tests to ensure higher regularization values results in lower weights (high regularization -> weights close to zero).

Pull request to add tests for issue #133

### DIFF
--- a/tests/test_regularization.py
+++ b/tests/test_regularization.py
@@ -1,7 +1,5 @@
 #!/usr/bin/env python3
 
-import os
-import pathlib
 import random
 
 import numpy as np
@@ -19,10 +17,6 @@ random.seed(0)
 np.random.seed(0)
 
 # Variables
-DIR = pathlib.Path(__file__).parent.parent.absolute()
-DATA_DIR = os.path.join(DIR, "tests", "test-data")
-PEYTON_FILE = os.path.join(DATA_DIR, "wp_log_peyton_manning.csv")
-
 EPOCHS = 10
 BATCH_SIZE_HOLIDAYS = 32
 BATCH_SIZE_EVENTS = 3
--- a/tests/test_regularization.py
+++ b/tests/test_regularization.py
@@ -21,7 +21,7 @@ EPOCHS = 10
 BATCH_SIZE_HOLIDAYS = 32
 BATCH_SIZE_EVENTS = 1
 LEARNING_RATE = 0.1
-REGULARIZATION = 0.02
+REGULARIZATION = 0.01
 # Map holiday name to a y value for dataset generation
 Y_HOLIDAYS_OVERRIDE = {
     "Washington's Birthday": 10,
--- a/tests/test_regularization.py
+++ b/tests/test_regularization.py
@@ -24,10 +24,10 @@ DATA_DIR = os.path.join(DIR, "tests", "test-data")
 PEYTON_FILE = os.path.join(DATA_DIR, "wp_log_peyton_manning.csv")
 
 NROWS = 100
-EPOCHS = 50
-BATCH_SIZE = 64
-LR = 1.0
-REGULARIZATION = 1e20
+EPOCHS = 10
+BATCH_SIZE = 32
+LEARNING_RATE = 0.1
+REGULARIZATION = 10
 
 
 def test_reg_func_abs():
@@ -48,9 +48,13 @@ def test_regularization_holidays():
     df = df_utils.check_dataframe(df, check_y=False)
 
     m = NeuralProphet(
+        yearly_seasonality=False,
+        weekly_seasonality=False,
+        daily_seasonality=False,
+        growth="off",
         epochs=EPOCHS,
         batch_size=BATCH_SIZE,
-        learning_rate=LR,
+        learning_rate=LEARNING_RATE,
     )
     m = m.add_country_holidays("US", regularization=REGULARIZATION)
     m.fit(df, freq="D")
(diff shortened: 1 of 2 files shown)


ourownstory:
Looking good!

Do you think we could do with lower epochs (like 5) and lower regularization (like 10)?

noxan:
Thanks for the feedback. I did give it a try with `epoch=5` and `regularization=10`. The problem with so few epochs is that the holiday weights do not take reliable values. Here is a Jupyter Notebook on Colab to see the results, just toggle the line with the holiday setup with and without regularization and a different number of epochs. Also added a synthetic dataset with `y=1` for all regular days and `y=1000` for all holidays, which might be handy for the unit tests.

https://colab.research.google.com/drive/1Z2KRiFvnL33RqYKLovhGmOodZydIfZ_y?usp=sharing

ourownstory:
> I did give it a try with `epoch=5` and `regularization=10`. The problem with so few epochs is that the holiday weights do not take reliable values. 

What do you suggest?


(If we merge events and holidays modeling, we could possibly just test for events regularization, as it's the same mechanism)

noxan:
Found a way to make it work with `epoch=10` using tuned hyperparameters and by reducing the batch size for events with the new artificial dataset.

Removing the second set of tests once #712 is done sounds like a great idea, meanwhile, I recommend keeping both.

noxan:
@ourownstory PR is ready to be reviewed. I've added holidays and events with regular values (therefore meaningless entries) and extended the test cases to expect low weights for those regular entries, as discussed earlier today.

tests/test_regularization.py
```diff
@@ -0,0 +1,170 @@
...
+# Variables
+EPOCHS = 10
+BATCH_SIZE_HOLIDAYS = 32
+BATCH_SIZE_EVENTS = 1
+LEARNING_RATE = 0.1
+REGULARIZATION = 10
+# Map holiday name to a y value for dataset generation
+Y_HOLIDAYS_OVERRIDE = {
+    "Washington's Birthday": 1,
+    "Labor Day": 5,
```
ourownstory, line 28: for simplicity it may be better to keep all override values the same.
I would suggest to have: 
* override events: 10x base value
* events: 100x base value
like this we would expect the overrides to have a weight of 0.1 and normal events 1.0
regularization should then reduce the overrides to <0.1 and normal >0.5

tests/test_regularization.py
```diff
@@ -0,0 +1,170 @@
...
+    m = m.add_country_holidays("US", regularization=REGULARIZATION)
+    m.fit(df, freq="D")
+
+    for country_holiday in m.country_holidays_config.holiday_names:
+        event_params = m.model.get_event_weights(country_holiday)
+        weight_list = [param.detach().numpy() for _, param in event_params.items()]
+        if country_holiday in Y_HOLIDAYS_OVERRIDE.keys():
+            assert weight_list[0] <= 0.1
+        else:
+            assert weight_list[0] <= 0.5
```
ourownstory, line 71: When regularizing, we still want to have the meaningful weights to not be reduced too much.
I think that this ought to be: `assert weight_list[0] > 0.5`?

tests/test_regularization.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+def test_regularization_holidays_disabled():
```
ourownstory, line 74: I think we can skip this test if you agree with my comment above .

tests/test_regularization.py
```diff
@@ -0,0 +1,170 @@
...
+    history_df = m.create_df_with_events(df, events_df)
+    m.fit(history_df, freq="D")
+
+    for index, event in enumerate(events):
+        weight_list = m.model.get_event_weights("event_%i" % index)
+        for _, param in weight_list.items():
+            if event in Y_EVENTS_OVERRIDE.keys():
+                assert param.detach().numpy() <= 0.1
+            else:
+                assert param.detach().numpy() <= 0.5
```
ourownstory, line 133: see comment for `test_regularization_holidays`

tests/test_regularization.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+def test_regularization_events_disabled():
```
ourownstory, line 136: see comment for `test_regularization_holidays_disabled`

ourownstory (changes requested):
Looking good!
See my two comments - we should flick an inequality check and can skip two tests.

tests/test_regularization.py
noxan: The original issue reported (#133) was that weights are not reduced to zero with regularization enabled. Happy to adjust the test case, but then we cannot be sure that regularization does actually reduce the weights in a meaningful way anymore. That's also why I added two test cases for each component (events and holidays), to make sure regularization has a meaningful impact on weights. Let me know your thoughts.

To ensure we still have meaningful weights in contrast to the meaningless weights we could add a check for the meaningful ones to have `weight_list[0] > 0.1`?

noxan:
@ourownstory all comments resolved as previously discussed, hope it's good to go, looking forward to your feedback.

ourownstory (approved):

ourownstory:
@noxan LGTM!

The only open question I have is: What prompted the regularization to be reduced from 10 to 0.02?
Maybe we should look into standardizing the scale of the regularization parameter to be more consistent.

noxan:
Had to scale regularization down to make sure the meaningful events keep their weight, seems like the regularization scale scales with the number of events and holidays, I will create a follow-up issue to track this further.

Final change: reduced regularization from `0.02` to `0.01` to add more safety for randomness in the tests.

PR #715, opened 2022-08-16
Uncertainty UI Enhancements

Kevin-Chen0:
This PR contains the following changes to Quantile Regression:

1. Set the default value of `quantiles` to `None` instead of `[0.5]`. Config's Train module will automatically set it up as `[0.5]` if `None` behind the scenes.
2. Add new param `confidence_interval`, which sets `quantiles` as `[confidence_interval/2, 1 - confidence_interval/2]`, with default value as `None`. This is a check that prevents both the `confidence_interval` and `quantiles` from containing values.


neuralprophet/configure.py
```diff
@@ -98,19 +99,33 @@ class Train:
...
         assert isinstance(self.quantiles, list), "Quantiles must be in a list format, not None or scalar."
 
-        # check for empty list
-        if len(self.quantiles) == 0:
-            raise ValueError("Please specify some quantile to estimate uncertainty")
+        # check if quantiles contain 0.5 and remove if so, as it will be inserted again as first index
+        if 0.5 in self.quantiles:
+            self.quantiles.remove(0.5)
 
         # check if quantiles are float values in (0, 1)
```
ourownstory, line 124: I think you already check for this above?

neuralprophet/configure.py
```diff
@@ -98,19 +99,33 @@ class Train:
...
+            self.quantiles = []
+
         # assert quantiles is a list type
         assert isinstance(self.quantiles, list), "Quantiles must be in a list format, not None or scalar."
 
-        # check for empty list
-        if len(self.quantiles) == 0:
-            raise ValueError("Please specify some quantile to estimate uncertainty")
+        # check if quantiles contain 0.5 and remove if so, as it will be inserted again as first index
+        if 0.5 in self.quantiles:
```
ourownstory, line 121: maybe we should use `math.isclose` to compare float numbers?

neuralprophet/forecaster.py
```diff
@@ -293,7 +293,8 @@ def __init__(
         optimizer="AdamW",
         newer_samples_weight=2,
         newer_samples_start=0.0,
-        quantiles=[0.5],
+        confidence_interval=None,
+        quantiles=None,
```
ourownstory, line 297: Let's also Introduce a new parameter: `uncertainty_estimation=”quantile_regression”`

ourownstory (changes requested):
Looking good! I just have three minor comments that should be quick to resolve.

neuralprophet/configure.py
Kevin-Chen0: No, this assertion is to check for any quantile that is equal to 0.5 so it can be removed. This is to prevent any duplicates as 0.5 is always automatically added to the first of the quantile list.

neuralprophet/configure.py
Kevin-Chen0: Maybe later on if this is needed.

neuralprophet/forecaster.py
Kevin-Chen0: This is done.

neuralprophet/forecaster.py
```diff
@@ -293,7 +301,9 @@ def __init__(
         optimizer="AdamW",
         newer_samples_weight=2,
         newer_samples_start=0.0,
-        quantiles=[0.5],
+        uncertainty_estimation="auto",
```
ourownstory, line 304: uncertainty_method

Kevin-Chen0: done

neuralprophet/forecaster.py
Kevin-Chen0: Renamed it to `uncertainty_method`.

neuralprophet/configure.py
Kevin-Chen0: This is done:

```
# check if quantiles contain 0.5 or close to 0.5, remove if so as 0.5 will be inserted again as first index
self.quantiles = [quantile for quantile in self.quantiles if not math.isclose(0.5, quantile)]
```

neuralprophet/configure.py
```diff
@@ -140,6 +126,39 @@ def __post_init__(self):
         if len(self.quantiles) > 1:
             self.loss_func = PinballLoss(loss_func=self.loss_func, quantiles=self.quantiles)
 
+    def set_quantiles(self):
+        # assert either prediction interval or quantiles is None, or both are None
+        assert self.prediction_interval is None or self.quantiles is None, "Prediction interval and quantiles cannot both be populated, one or both must be None."
+        if self.uncertainty_method.lower() in ["quantile_regression", "quantile regression", "qr"]:
+            # assert prediction interval is None and quantiles is a list
+            assert self.prediction_interval is None and isinstance(self.quantiles, list), "Prediction interval must be None while quantiles must be a list. Otherwise , set uncertainty estimation param to 'auto'"
```
ourownstory, line 134: ..., set uncertainty_method

neuralprophet/configure.py
```diff
@@ -140,6 +126,39 @@ def __post_init__(self):
...
+        elif self.uncertainty_method.lower() in ["conformal_prediction", "conformal prediction", "cp"]:
+            # assert prediction interval is a float and quantiles is None
+            assert isinstance(self.prediction_interval, float) and self.quantiles is None, "Prediction interval must be a float while quantiles must be None. Otherwise , set uncertainty estimation param to 'auto'"
```
ourownstory, line 137: ..., set uncertainty_method

neuralprophet/configure.py
```diff
@@ -140,6 +126,39 @@ def __post_init__(self):
...
+        elif self.uncertainty_method.lower() in ["auto", "a"]:
+            # assert that prediction interval is a float between (0, 1) if not None, then use that to create the quantiles
+            if self.prediction_interval is not None:
+                assert isinstance(self.prediction_interval, float) and (0 < self.prediction_interval < 1), \
+                    "The prediction interval specified needs to be a float in-between (0, 1)."
+                alpha = 1 - self.prediction_interval
+                self.quantiles = [alpha/2, 1 - alpha/2]
+        else:
+            raise ValueError("The only valid uncertainty_method options are 'auto', 'quantile_regression', or 'conformal_prediction'.")
```
ourownstory, line 146: Let's leave conformal_prediction out of the message until it's available

neuralprophet/configure.py
Kevin-Chen0: I don't follow.

neuralprophet/configure.py
```diff
@@ -140,6 +126,39 @@ def __post_init__(self):
...
+        assert self.prediction_interval is None or self.quantiles is None, "Prediction interval and quantiles cannot both be populated, one or both must be None."
+        if self.uncertainty_method.lower() in ["quantile_regression", "quantile regression", "qr"]:
+            # assert prediction interval is None and quantiles is a list
+            assert self.prediction_interval is None and isinstance(self.quantiles, list), "Prediction interval must be None while quantiles must be a list. Otherwise , set uncertainty estimation param to 'auto'"
+        elif self.uncertainty_method.lower() in ["conformal_prediction", "conformal prediction", "cp"]:
+            # assert prediction interval is a float and quantiles is None
+            assert isinstance(self.prediction_interval, float) and self.quantiles is None, "Prediction interval must be a float while quantiles must be None. Otherwise , set uncertainty estimation param to 'auto'"
+        elif self.uncertainty_method.lower() in ["auto", "a"]:
+            # assert that prediction interval is a float between (0, 1) if not None, then use that to create the quantiles
+            if self.prediction_interval is not None:
```
ourownstory, line 140: If quantiles are provided, this would overwrite them. I think quantiles are more explicit, and thus they should take precedence.
I suggest to change this line to:
`if self.prediction_interval is not None and self.quantiles is None:`
-> also, after setting quantiles, set `self.prediction_interval = None`

and additionally
`elif self.prediction_interval is not None and self.quantiles is None:`
-> log.warn("Ignoring prediction_interval and using specified quantiles for uncertainty estimation")
and also: set `self.prediction_interval = None`

Kevin-Chen0: The first assertion statement prevents both `self.prediction_interval` and `self.quantiles` from beings values other than `None`.

`assert self.prediction_interval is None or self.quantiles is None, "Prediction interval and quantiles cannot both be populated, one or both must be None."`

Nevertheless, for clarity, I'll still change this line to `if self.prediction_interval is not None and self.quantiles is None:` and set `self.prediction_interval = None` after.

neuralprophet/configure.py
Kevin-Chen0: Done

neuralprophet/configure.py
Kevin-Chen0: I pushed, which incorporated your suggestions for the first half. However, I'm still unsure about what you meant regarding the part after "additionally".

neuralprophet/configure.py
ourownstory: maybe change the message to:
"When uncertainty_method is 'quantile_regression', specify quantiles as list and do not set prediction_interval."
(to make it clearer to which parameters the error refers)

neuralprophet/configure.py
ourownstory: the part of after _additionally_ was refering to handling the case when we would not raise an error if both `prediction_interval` and `quantiles` are populated.

ourownstory: But we can raise an error in this case as you configured now, no need to change.

neuralprophet/forecaster.py
```diff
@@ -212,7 +212,15 @@ class NeuralProphet:
...
-        quantiles: list, default [0.5]
+        uncertainty_method : str, default ``auto``
+            Specifies the type of uncertainty estimation technique that is being deployed 
+        
+            Options
+                * (default) ``auto``: Automatically infers the uncertainty estimation technique based on the prediction interval or quantiles params.  
+                * ``quantile_regression``: Requires the quantiles to be specified.
+        prediction_interval : float, default None 
+            Width of the uncertainty or confidence intervals provided for the forecast. Must be between (0, 1).
+        quantiles : list, default None
```
ourownstory, line 223: We could we add a mini example, similar to:
 https://github.com/ourownstory/neural_prophet/pull/746/files

neuralprophet/forecaster.py
```diff
@@ -217,11 +217,18 @@ class NeuralProphet:
         
             Options
                 * (default) ``auto``: Automatically infers the uncertainty estimation technique based on the prediction interval or quantiles params.  
-                * ``quantile_regression``: Requires the quantiles to be specified.
+                * ``quantile_regression``: Requires the quantiles to be specified while leaving prediction_interval as None.
+                * ``conformal_prediction``: Requires the prediction_interval to be while leaving quantiles as None. (WIP)
```
ourownstory, line 221: let's delete this line in this PR and introduce it in th CP PR :)

neuralprophet/forecaster.py
```diff
@@ -217,11 +217,18 @@ class NeuralProphet:
...
+
+            Examples
+            --------
+            >>> from neuralprophet import NeuralProphet
+            >>> m = NeuralProphet(uncertainty_method="quantile_regression", quantiles=[0.5, 0.95])
```
ourownstory, line 226: did you mean? `quantiles=[0.05, 0.95]`

PR #718, opened 2022-08-16
add hint to add_country_holidays that previous value will be overridden

### LINKED ISSUE #689: add_country_holidays does replace prior holiday values in contrast to its name
Calling `add_country_holidays` does override any previously defined holidays. Meaning repeated calls of `add_country_holidays` with different countries do only result in the holidays of the final call (see example below).

Looking up the implementation/documentation of prophet their function behaves the same day, so either we should
1) add the same hint in documentation (see [reference](https://github.com/facebook/prophet/blob/10310ceb2da05837a198db6714d658a1e0a32478/python/prophet/forecaster.py#L714)), possibly throw an exception/warning in case people call it multiple times
2) differ from the behavior of prophet and actually allow for multiple countries
3) renaming the function to `set_country_holidays` to clarify its behavior.

Would be curious to hear if only I got confused and which direction forward is best.

Example from an integration test with multiple `add_country_holidays` calls: https://github.com/ourownstory/neural_prophet/blob/main/tests/test_integration.py#L429-L434

noxan:
fix #689

### DIFF
--- a/neuralprophet/forecaster.py
+++ b/neuralprophet/forecaster.py
@@ -555,6 +555,10 @@ class NeuralProphet:
         """
         if self.fitted:
             raise Exception("Country must be specified prior to model fitting.")
+        if self.country_holidays_config:
+            log.warning(
+                "Country holidays can only be added for a single country. Previous country holidays were overridden."
+            )
 
         if regularization is not None:
             if regularization < 0:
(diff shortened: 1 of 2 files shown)


ourownstory:
@noxan Thank you for adding this!
Might you be able to also add a log.warning message as you suggested in point 1 in #689?

ourownstory (approved):

Issue #744, opened 2022-08-30
[bug] drop_missing=True causes error if m.predict() is called without extended df


judussoari:
**Describe the bug**

If `drop_missing` is activated, the code executes as long as we are predicting into the unknown future (i.e. a future df is created using `m.make_future_dataframe()`). 
If the df is not extended into the unknown future (i.e. `m.make_future_dataframe() `is not used), we run into a dimension error: 
`ValueError: operands could not be broadcast together with shapes (520,) (542,)` 

**To Reproduce**

```
from neuralprophet import NeuralProphet, set_log_level
set_log_level("ERROR")
data_location = "https://raw.githubusercontent.com/ourownstory/neuralprophet-data/main/datasets/"
df = pd.read_csv(data_location + "wp_log_peyton_manning.csv")

m = NeuralProphet(
n_lags=12,
n_forecasts=1,
impute_missing=True,
drop_missing=True,
)

# introduce large window of NaN values, from which samples will be dropped after imputation
df["y"][100:131] = np.nan
metrics = m.fit(df, freq="D", validation_df=None)
forecast = m.predict(df=df)
```

judussoari:
Fixed in #786

PR #747, opened 2022-09-01
[bug-fix] Enable predictions into known and unknown future in case of remaining NaN windows after imputation

judussoari:
Referring to Issue #744 

**Problem**
1. The code ran into an Error whenever the user opted to drop missing values that remained after imputation, with no extended df (i.e. no prediction into the unknown future)
2.  For predictions into the unknown future (i.e. using `m.make_future_dataframe()`) with `n_lags=0`, the resulting future dataframe contains `n_historic_predictions` valid values, and `periods` invalid (None/NaN) values. Feeding this df into `m._create_dataset()` either causes an Error (if `drop_missing=False`) or drops all the `periods` in the df (if `drop_missing=True`), which is not desired.

**Solution**
1. Similar to samples (inputs/targets) being dropped from the data, the same datapoints (identifiable through sample indices) must be dropped from the original/scaled df as well, in order to create the df with predictions yhat1, yhat2... 
2. For predicting into the unknown future, the NaN values at the end of `df_future` are excluded from the data drop process. This is enabled through a new kwarg `predict_steps` that is either equal to `n_forecasts` (case of forecast into known future) or `periods` (into unknown future)


judussoari:
> Thank you @judussoari ! First quick question: Does your solution to 2 mean that `n_historic_predictions` is ignored or are you cropping it accordingly after appending the `periods` of None values?

Hey @ourownstory, `n_historic_predictions` is not ignored. Exactly like for `n_lags>0` before, the future df now consists of `n_historic_predictions` and `periods` NaN values. However, periods is now **always** (i.e. `n_lags>=0` instead of `n_lags>0` only) set to `m.n_forecasts`.

neuralprophet/forecaster.py
```diff
@@ -1922,6 +1925,8 @@ def __handle_missing_data(self, df, freq, predicting):
         data_columns = []
         if self.max_lags > 0:
             data_columns.append("y")
+        if self.max_lags == 0 and predicting:  # case of n_lags = 0 and predicting into the known future
```
ourownstory, line 1928: drop

neuralprophet/forecaster.py
```diff
@@ -1922,6 +1925,8 @@ def __handle_missing_data(self, df, freq, predicting):
         data_columns = []
         if self.max_lags > 0:
```
ourownstory, line 1926: n_lags

neuralprophet/time_dataset.py
```diff
@@ -368,11 +373,10 @@ def _stride_lagged_features(df_col_name, feature_dims):
 
         inputs["events"] = events
 
-    if predict_mode:
```
ourownstory, line 371: let's keep this

ourownstory (commented):
When predicting (only) and dropping nan, only drop values related to lags, not n_forecasts

neuralprophet/time_dataset.py
```diff
@@ -75,6 +75,10 @@ def drop_nan_after_init(self, drop_missing):
                 whether to automatically drop missing samples from the data
         """
         nan_idx = []
+        nan_at_end = 0
```
ourownstory, line 78: let's get this as a kwarg `predict_steps` (n_forecasts or periods)

ourownstory:
updated in #786

PR #751, opened 2022-09-02
[Feature] Include quantiles in get_latest_forecast() output df 

judussoari:
output df of utils.fcst_df_to_last_forecast() now contains quantile columns if quantiles are specified in the model

quantile columns are generated in the same way as the yhat columns

### DIFF
--- a/neuralprophet/forecaster.py
+++ b/neuralprophet/forecaster.py
@@ -720,6 +720,24 @@ class NeuralProphet:
         forecast = pd.DataFrame()
         for df_name, df_i in df.groupby("ID"):
             dates, predicted, components = self._predict_raw(df_i, df_name, include_components=decompose)
+            if self.config_missing.drop_missing and self.max_lags > 0:
+                # drop NaN windows (similar to lags/targets) in df, but not the NaNs that were inserted for prediction at the end
+                while np.isnan(np.array(df_i["y"][: -self.n_forecasts])).any():
+                    window = []
+                    all_nan_idx = (
+                        df_i[: -self.n_forecasts].loc[np.isnan(np.array(df_i["y"][: -self.n_forecasts]))].index
+                    )
+                    for i in range(len(all_nan_idx)):
+                        window.append(all_nan_idx[i])
+                        if all_nan_idx.max() == all_nan_idx[i]:
+                            break
+                        if all_nan_idx[i + 1] - all_nan_idx[i] > 1:
+                            break
+                    df_i = (
+                        df_i.drop(df_i.index[(window[0] - (self.max_lags + self.n_forecasts - 1)) : window[-1] + 1])
+                        .reset_index()
+                        .drop("index", axis=1)
+                    )
             if raw:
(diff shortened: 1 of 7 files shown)


neuralprophet/utils.py
```diff
@@ -490,7 +490,7 @@ def print_epoch_metrics(metrics, val_metrics=None, e=0):
     return metrics_string
 
 
-def fcst_df_to_last_forecast(fcst, n_last=1):
+def fcst_df_to_last_forecast(fcst, quantiles, n_last=1):
     """Converts from line-per-lag to line-per-forecast.
```
ourownstory, line 494: please expand on the args in a docstring

neuralprophet/time_dataset.py
```diff
@@ -383,7 +385,7 @@ def _stride_lagged_features(df_col_name, feature_dims):
             tabularized_input_shapes_str += ("    {} {} \n").format(key, value.shape)
     log.debug("Tabularized inputs shapes: \n{}".format(tabularized_input_shapes_str))
 
-    return inputs, targets, config_missing.drop_missing
+    return inputs, targets, config_missing.drop_missing, n_forecasts
```
ourownstory, line 388: why do we need to return n_forecasts here?

neuralprophet/plot_forecast.py
```diff
@@ -81,7 +81,7 @@ def plot(
     else:
         fig = ax.get_figure()
     ds = fcst["ds"].dt.to_pydatetime()
-    yhat_col_names = [col_name for col_name in fcst.columns if "yhat" in col_name]
+    yhat_col_names = [col_name for col_name in fcst.columns if "yhat" in col_name and "%" not in col_name]
```
ourownstory, line 84: will this not affect regular plotting of quantiles?

neuralprophet/forecaster.py
```diff
@@ -720,6 +720,24 @@ def predict(self, df, decompose=True, raw=False):
         forecast = pd.DataFrame()
         for df_name, df_i in df.groupby("ID"):
             dates, predicted, components = self._predict_raw(df_i, df_name, include_components=decompose)
+            if self.config_missing.drop_missing and self.max_lags > 0:
```
ourownstory, line 723: could we make this into a function?
Why do we need to call this in predict?

neuralprophet/utils.py
judussoari: done! Pls see PR #764 as this PR will be closed

neuralprophet/time_dataset.py
judussoari: I found a solution without returning n_forecasts. pls see PR #747 as this PR will be closed

neuralprophet/plot_forecast.py
judussoari: You are right! Pls see PR #764 where I found a better solution

neuralprophet/forecaster.py
judussoari: Done! Pls see PR #747

Issue #684, opened 2022-07-12
[bug] np.quantile not compatible with python < 3.8


Mckiler:
I found some error on energy_solar_pv.ipynb

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
[<ipython-input-31-bfeb12738ca9>](https://localhost:8080/#) in <module>()
      1 m = NeuralProphet()
----> 2 metrics = m.fit(df_train)
      3 forecast = m.predict(df_test)

6 frames
[/usr/local/lib/python3.7/dist-packages/neuralprophet/df_utils.py](https://localhost:8080/#) in get_normalization_params(array, norm_type)
    340     if norm_type == "soft":
    341         lowest = np.min(non_nan_array)
--> 342         q95 = np.quantile(non_nan_array, 0.95, method="higher")
    343         width = q95 - lowest
    344         if math.isclose(width, 0):

<__array_function__ internals> in quantile(*args, **kwargs)

TypeError: _quantile_dispatcher() got an unexpected keyword argument 'method'

rlxm7100:
I have same issue too.

FedericoGarza:
Hi @Mckiler, @rlxm7100!

I had the same problem. The problem arises from the version of `NumPy` used. The `method` argument of the `np.quantile` function is new as of [`NumPy` version 1.22.1](https://github.com/ourownstory/neural_prophet/issues/684), but that version is only compatible with [python>=3.8](https://numpy.org/devdocs/release/1.22.1-notes.html). 

The solution is to use this version of python (3.8) and install `NumPy` and `neuralprophet`.

@ourownstory It could be useful to include this information in the colab notebooks since the default version in colab is python 3.7.

PR #752, opened 2022-09-02
[bugfix] Addressed np.quantile issue

karl-richter:
In #684, users addressed that the `np.quantile` function is incompatible with any Python version lower than 3.8.
This is due to a change in the Numpy api introduced with version 1.22.0, however, this version is not compatible with any older Python version.

In the [Numpy docs](https://numpy.org/doc/stable/reference/generated/numpy.quantile.html), the developers state that they renamed the `method` function argument (previously: `interpolation`) and added argument options. Since the option we are using (`method="higher"`) has previously been available, the simple solution I propose is to wrap the statement with a conditional that switches between the old and the new function call based on the installed Numpy version. The `np.quantile` is only called two times in the same function within the whole repo, this solution adds only a minor technical overhead.

Let me know what you are thinking!

```python
if np.__version__ >= "1.22.0":
      q95 = np.quantile(non_nan_array, 0.95, method="higher")
else:
      q95 = np.quantile(non_nan_array, 0.95, interpolation="higher")
```

### DIFF
--- a/neuralprophet/df_utils.py
+++ b/neuralprophet/df_utils.py
@@ -339,7 +339,10 @@ def get_normalization_params(array, norm_type):
     non_nan_array = array[~np.isnan(array)]
     if norm_type == "soft":
         lowest = np.min(non_nan_array)
-        q95 = np.quantile(non_nan_array, 0.95, method="higher")
+        if np.__version__ >= "1.22.0":
+            q95 = np.quantile(non_nan_array, 0.95, method="higher")
+        else:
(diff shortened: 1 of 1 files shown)


ourownstory:
@karl-richter Thank you, that looks good to me. 
Looking closer into the docs of np.quantile, I wonder if we should just not specify the method. The default is 'linear' both before and after 1.22. As we work exclusively with numeric values, this should in most cases lead to similar results as the 'higher' interpolation. If we were to work with categoric values, we would likely anyhow need to use a different data normalization method alltogether.

PR #756, opened 2022-09-05
Change str.format() to f-string

YoussefKafa:
**issue: #755**
Change str.format() to f-string throughout the neuralprophet source code.

PR #757, opened 2022-09-06
Changed all .format() to fstring

Rugbeatuno:
I changed all formatted strings to fstring, with the exception of instances in 
**tutorials/feature-use/lagged_covariates_energy_ercot.ipynb**
and 3 other instances that were commented out.

This is my first PR ever so I really hope I don't break everything, I did my best to check everything I changed but with over 300 changes one could have slipped past.


Kevin-Chen0:
Thanks for all of your effort @Rugbeatuno! I'll review it and get back to you.

ourownstory:
@Rugbeatuno Thank you so much for your contribution!

It seems like you and @YoussefKafa both worked on the same issue in parallel, see #756.
As Youssef seems to have opened his PR just slightly before you, we will first attempt at merging his PR, but there may still be areas that you covered which he missed. Once his is merged, we will see what of the diff of your branch to main remains and then you can add that. 

To avoid these situations, I strongly recommend opening a PR when you *start* to work on an issue, (don't wait until done), so others can see that work is underway.

PR #764, opened 2022-09-08
[Feature] Include quantiles in get_latest_forecast() output df

judussoari:
*output df of utils.fcst_df_to_last_forecast() now contains quantile columns if quantiles are specified in the model

*quantile columns are generated in the same way as the yhat columns

*after enabling quantiles in get_latest_forecast(), the quantile names need to be ignored in case of highlight_forecasts is not none & line_per_origin = True


tests/test_uncertainty.py
```diff
@@ -64,6 +64,9 @@ def test_uncertainty_estimation_plot():
     m.plot_parameters()
     if PLOT:
         plt.show()
+    m.plot_last_forecast(forecast, include_previous_forecasts=10, plot_history_data=True)
```
ourownstory, line 67: fig1 = m. plot_...

ourownstory:
updated in #785

PR #766, opened 2022-09-13
remove print statement from forecaster.py

noxan:
Removed some left over debug print statement. In case we want to keep it, better move it to logging.


ourownstory:
Seems like this was introduced in #715 @Kevin-Chen0

PR #773, opened 2022-09-17
[enhancement] Change str.format() to f-string

Kevin-Chen0:


Kevin-Chen0:
Hey @ourownstory @karl-richter, do you know what could be causing the code coverage to be so low? I am just changing the strings to f-strings, right? So it shouldn't be affecting any of the functionalities.

ourownstory:
> Hey @ourownstory @karl-richter, do you know what could be causing the code coverage to be so low? I am just changing the strings to f-strings, right? So it shouldn't be affecting any of the functionalities.

Hey @Kevin-Chen0 It is because most of the strings that are formatted are from Error and log printouts. We currently do not have test cases for everything that could go wrong to throw a warning, but mostly just test for things that should work to work.
So, I as you are not editing fuctional code here, but just mainly warning/error messages, the low coverage is fine from my point of view.

ourownstory (approved):
LGTM. 
Did you run black over these files?

PR #551, opened 2022-02-18
Bugfix: return correctly scaled multiplicative seasonality component

denyslazarenko:
Resolve issue: https://github.com/ourownstory/neural_prophet/issues/429
Used hint:
trend =  trend(scaled)*y_scale + y_shift
season_x = season_x(scaled) *y_scale
season_x_m = season_x_m(scaled)*trend(scaled)*y_scale
season_x_m = season_x_m(scaled)*(trend - y_shift)

Issue #429, opened 2021-10-15
Multiplicative Seasonality fitted correctly but returned wrong


ourownstory:
### Discussed in https://github.com/ourownstory/neural_prophet/discussions/303

<div type='discussions-op-text'>

<sup>Originally posted by **sgheinze** March 26, 2021</sup>
I'm hoping someone can help me understand the multiplicative model used in NeuralProphet and how it relates to values in the forecast DataFrame.

I've (perhaps incorrectly) been assuming that the NeuralProphet model is the same as the Prophet model, which according to this post in the Prophet GitHub (https://github.com/facebook/prophet/issues/1329#issuecomment-588536185) and this kaggle source (https://www.kaggle.com/smakarychev/what-is-facebook-prophet-forecasting-model), is:
`Yhat = trend * (1 + multiplicative terms) + additive terms`
which for models without additive terms reduces to:
`Yhat = trend * (1 + multiplicative terms)`

However, when I fit a simple multiplicative model to some toy data with only a single multiplicative seasonal term, I am unable to verify that equation:
```
model = NeuralProphet(n_forecasts=1, seasonality_mode="multiplicative", growth="off", weekly_seasonality=True, yearly_seasonality=False)
metrics = model.fit(df, freq="D")
future = model.make_future_dataframe(df, periods=1)
forecast = model.predict(future)
forecast.head()

	ds	          y	       yhat1  residual1	     trend	 season_weekly
0	2021-03-18	None	27814.070312	NaN	30103.738281	-0.064573
```
As you can see in the forecast DataFrame above, `yhat1 != trend * (1 + season_weekly)` for this particular forecast.

Where am I going wrong?</div>

PR #791, opened 2022-09-24
[bugfix] fix re-scaling of multiplicative components in forecast df

LeonieFreisinger:
refer to issue #429 
makes PR #551 unnecessary

**Problem**
Multiplicative components are not output correctly in the forecast df due to being computed incorrectly. 

**Solution**
The multiplicative components are not re-scaled properly. Therefore, the re-scaling of all multiplicative components is added in predict_raw().


LeonieFreisinger:
@ourownstory @Kevin-Chen0 Feel free to review the implementation of the proper re-scaling of the multiplicative components. Thanks a lot!

PR #792, opened 2022-09-26
Lightning Migration

karl-richter:
### v0 Todos
- [x] Change TimeNet parent from PyTorch Module to PyTorch Lightning
- [x] Add PyTorch Lightning functions to TimeNet (eg. training_step)
- [x] Add Lightning Trainer in forecaster
- [x] Change train and predict functions in forecaster to use the Lightning Trainer
- [x] Change the prediction logic in forecaster.py 
- [x] Handle `minimal` mode, eg. `_train_minimal(...)`
- [x] Return metrics in `_train(...)`
- [x] Define epochs, batch, lr (etc.) correctly in trainer (when not provided) + move to fit() method

### v1 Todos
- [x] Remove outdated code
- [x] Change model saving and loading from checkpoint [Checkpointing](https://pytorch-lightning.readthedocs.io/en/stable/common/checkpointing_basic.html)
- [x] Add support for all Lightning loggers (eg. Tensorboard)
- [x] Learning rate range finder
- [x] Use correct batch_size
- [x] Regularisation loss fixed
- [x] Pass denomralization as function
- [ ]  Support the `self.metrics.add_specific_target` function when `highlight_forecast_step_n` is defined

### v2 Todos (separate PRs)
- [ ] Continue training mode #828 
- [ ] Support additional training Loggers (eg. TensorBoardLogger) #830 
- [ ] GPU support
- [ ] Early stopping

### Changes 
- Use the Lightning logic for training, evaluation and prediction (removes training logic from forecaster.py) [Docs](https://pytorch-lightning.readthedocs.io/en/latest/)
- Use torchmetrics to calculate metrics (instead of using the custom metrics object) [Docs](https://torchmetrics.readthedocs.io/en/stable/)
- Use the Lightning built-in progress bar for training live loss (use the rich progress bar instead of the custom tqdm progress bar and the default lightning tqdm progress bar due to issues in Jupyter notebook) [Docs](https://pytorch-lightning.readthedocs.io/en/stable/common/progress_bar.html)
- Use a custom logger for Lightning to collect metrics in a dictionary (the default logger saves and checkpoints the model automatically, which we dont always want)
- Added support for early stopping using the Lightning loss monitor (with that we can close #289)
- Use the Lightning built-in Learning rate finder [Docs](https://pytorch-lightning.readthedocs.io/en/1.4.5/advanced/lr_finder.html)

**Before**
The training logic is contained in `forecaster.py`, manually iter through epochs and batches. The `train_epoch()` function directly calls `forward()` on the TimeNet model. Optimization happens manually in `train_epoch()`. 

```mermaid
flowchart LR
    subgraph TimeNet
    forward
    end
    subgraph NeuralProphet
    fit --> _train
    _train --> _train_epoch
    _train_epoch --> forward
    predict --> _predict_raw
    _predict_raw --> forward
    end
```

**After**
The training loop is abstracted using the Lightning training logic. Init a Lightning `Trainer` object that can run the training loop automatically. Calling the `fit()` method on the Lightning trainer executes the `training_step()` function of the model using the correct epoch and batch. Optimization and parameter updates happen automatically after the `training_step()` function. The whole training logic is abstracted away. Lightning provides useful tools such as a progress bar, a learning rate finder, early stopping, GPU support etc.

```mermaid
flowchart LR
    subgraph TimeNet
    configure_optimizers
    training_step --> forward
    predict_step --> forward
    end
    subgraph LightningTrainer
    tune --> configure_optimizers
    fit_[fit] --> training_step
    predict_[predict] --> predict_step
    end
    subgraph NeuralProphet
    fit --> _train
    _train --> tune
    _train --> fit_
     predict --> _predict_raw
     _predict_raw --> predict_
    end
```


neuralprophet/configure.py
```diff
@@ -161,19 +171,41 @@ def set_auto_batch_epoch(
...
-            div_factor=100.0,
-            final_div_factor=5000.0,
+    def set_optimizer(self):
+        """
+        Set the optimizer and optimizer args. If optimizer is a string, then it will be converted to the corresponding torch optimizer.
+        The optimizer is not initialized yet as this is done in configure_optimizers in TimeNet.
+        """
+        self.optimizer, self.optimizer_args = utils_torch.create_optimizer_from_config(
+            self.optimizer, self.optimizer_args
+        )
```
karl-richter, line 181: @ourownstory I would propose to move the logic from `create_optimizer_from_config` directly into the `set_optimzer` function since that's how we also handle it for `set_loss_func` and the other setter functions. What do you think?

neuralprophet/configure.py
```diff
@@ -161,19 +171,41 @@ def set_auto_batch_epoch(
...
+        """
+        Set the lr_finder_args.
+        This is the range of learning rates to test.
+        """
+        self.lr_finder_args.update(
+            {
+                "min_lr": 1e-7,
+                "max_lr": 100,
+                "num_training": 50 + int(np.log10(100 + dataset_size) * 25),
+            }
```
karl-richter, line 208: Currently this does not allow the user to overwrite the default `min_lr` etc., should we allow the user to change it or are we fine with keeping it static for now.

neuralprophet/logger.py
```diff
@@ -0,0 +1,54 @@
...
+from pytorch_lightning.utilities.distributed import rank_zero_only
+from pytorch_lightning.loggers.logger import rank_zero_experiment
+
+
+class MetricsLogger(Logger):
+    def __init__(self):
+        super().__init__()
+
+        self.history = collections.defaultdict(list)
+        self.checkpoint_path = None
```
karl-richter, line 14: Info: The default Lightning loggers do not maintain the provided logs in-memory but directly store them to a file. Since we want to access the metrics directly after training and not necessarily want to persist them, I defined a custom logger that maintains the logs in-memory. This allows us to access them right after training by calling `logger.history`

neuralprophet/metrics.py
```diff
@@ -4,10 +4,17 @@
...
 import logging
+import torchmetrics
 
 log = logging.getLogger("NP.metrics")
 
+METRICS = {
+    "MAE": torchmetrics.MeanAbsoluteError(),
+    "MSE": torchmetrics.MeanSquaredError(squared=True),
+    "RMSE": torchmetrics.MeanSquaredError(squared=False),
+}
```
karl-richter, line 15: Info: We use torchmetrics to log metrics during training. Torchmetrics was originally part of the Lightning package but moved into a seperate package to reduce complexity. It is maintained by the same authors.

neuralprophet/time_net.py
```diff
@@ -767,6 +830,165 @@ def compute_components(self, inputs):
...
+        # Calculate loss
+        loss, reg_loss = self.loss_func(inputs, predicted, targets)
+        # Metrics
+        if not self.minimal:
+            predicted_denorm = self.denormalize(predicted[:, :, 0])
+            target_denorm = self.denormalize(targets.squeeze(dim=2))
+            self.log_dict(self.metrics_train(predicted_denorm, target_denorm), **self.log_args)
+            self.log("Loss", loss, **self.log_args)
+            self.log("RegLoss", reg_loss, **self.log_args)
+        return loss
```
karl-richter, line 862: Info: The `_step` functions are used by Lightning to execute the training procedure. After the `training_step`, Lightning automatically calls the optimizer and learning rate scheduler to update the model parameters.

neuralprophet/time_net.py
```diff
@@ -767,6 +830,165 @@ def compute_components(self, inputs):
...
+                max_lr=self.learning_rate,
+                epochs=self.trainer.max_epochs,
+                steps_per_epoch=steps_per_epoch,
+                **self.config_train.scheduler_args,
+            ),
+            "name": "learning_rate",
+            "interval": "step",
+        }
+
+        return [optimizer], [lr_scheduler]
```
karl-richter, line 919: Info: The optimizer and scheduler are only initialized here, since `self.learning_rate` is accessed and tuned by the learning rate scheduler. Further, args such as the number of steps per epoch are only available at runtime in the trainer object and not before the model is trained.

neuralprophet/utils.py
```diff
@@ -1,35 +1,43 @@
...
-        >>> save(model, "test_save_model.np")
+        >>> save(forecaster, "test_save_model.np")
     """
-    torch.save(model, path)
+    # Remove non-serializable components (model, trainer, metrics)
+    for attr in ["metrics", "model", "trainer"]:
+        setattr(forecaster, attr, None)
+    # Add the model back in after saving (workaround for PyTorch Lightning)
+    forecaster.model = time_net.TimeNet.load_from_checkpoint(forecaster.metrics_logger.checkpoint_path)
+    torch.save(forecaster, path)
```
karl-richter, line 40: The Lightning objects cannot directly be serialized (because some tensors are still in grad-mode), thus they are removed from the NeuralProphet object. Workaround: After reading the model from a checkpoint however, we can make it part of the forecaster again.

PR #793, opened 2022-09-27
[Feature] Plot quantiles for very last forecast

judussoari:
-always plot last quantiles (before: only in case of highlight_forecast=None)

-fixed plot labels in plot_last_forecast()


judussoari:
I am not sure why the project codecov is lower because my changes are fully covered. As this is just an insignificant difference, I'd ask the reviewers to give their ok if everything else is fine!

ourownstory (changes requested):
Looks good, please just add one example plot to plotly tutorial.

judussoari:
Resolved in PR #829

PR #802, opened 2022-10-09
Uncertainty: Conformal Prediction V1

Kevin-Chen0:
Incorporate the first 2 split conformal prediction (SCP) methods to create prediction intervals for NeuralProphet:
- Absolute difference between y and y-hat (Naive)
- Conformalized Quantile Regression (CQR) - [Source](https://arxiv.org/abs/1905.03222)

Example code:

```
m = NeuralProphet(quantiles=[0.05, 0.95])
metrics = m.fit(train_df)
m.conformalize(calibration_df, alpha=0.1, method='cqr')  # Create the conformal prediction interval width q
forecast = m.predict(test_df)  # Automatically includes two predictive columns with +q and -q
m = m.highlight_nth_step_ahead_of_each_forecast(1)
m.plot(forecast)  # Automatically plots two predictive columns with +q and -q 
```

Use the [conformal_prediction_energy_hospital_load.ipynb](https://github.com/ourownstory/neural_prophet/blob/conformal-prediction/tutorials/feature-use/conformal_prediction_energy_hospital_load.ipynb) example notebook as test.


ourownstory:
@Kevin-Chen0 some quick feedback:
- Summary in the notebook seems no longer aligned with results?
- I think we can skip model 1 as it is nearly identical to m2?
- Consider reducing your regularization rates to 0.1 from 1.0 to avoid potential confounding issues.

Kevin-Chen0:
> @Kevin-Chen0 some quick feedback:
> 
> * Summary in the notebook seems no longer aligned with results?
> * I think we can skip model 1 as it is nearly identical to m2?
> * Consider reducing your regularization rates to 0.1 from 1.0 to avoid potential confounding issues.

* Will fix
* No, m1 is the default NP while m2 is MP with biased seasonality. I got feedback from a user who suggested creating a default NP model as a baseline, hence the current m1. Also, their interval width is quite different for the naive SCP.
* Do you mean the `ar_reg` for m3?

neuralprophet/forecaster.py
```diff
@@ -340,6 +341,7 @@ def __init__(
         # General
         self.name = "NeuralProphet"
         self.n_forecasts = n_forecasts
+        self.q_hats = []
```
ourownstory, line 344: As this variable is specific to CP, I think it should ideally not be a general class attribute.
Maybe we can make it a part of the uncertainty config? It is only estimated once and then retrieved, so the initialized config should be an OK place for it IMO.
(there, the name would also be less ambiguous, but maybe a slightly more verbose name may help, there, too)

neuralprophet/forecaster.py
```diff
@@ -811,6 +813,18 @@ def predict(self, df, decompose=True, raw=False):
             forecast = pd.concat((forecast, fcst), ignore_index=True)
         df = df_utils.return_df_in_original_format(forecast, received_ID_col, received_single_time_series)
         self.predict_steps = self.n_forecasts
+        # Conformal prediction interval with q
+        if self.q_hats:
+            if self.conformal_method == "naive":
+                df["yhat1 - qhat1"] = df["yhat1"] - self.q_hats[0]
```
ourownstory, line 819: I think it would be more understandable to the general user, if we keep uncertainty naming consistent and also just show one uncertainty value in the df. 
As we currently use `f"yhat1 {quantile_lo}%"` and  `f"yhat1 {quantile_hi}%"` I suggest we stick with those.

neuralprophet/forecaster.py
```diff
@@ -811,6 +813,18 @@ def predict(self, df, decompose=True, raw=False):
...
+                df["yhat1 + qhat1"] = df["yhat1"] + self.q_hats[0]
+            else:  # self.conformal_method == "cqr"
+                quantile_hi = str(max(self.config_train.quantiles) * 100)
+                quantile_lo = str(min(self.config_train.quantiles) * 100)
+                df[f"yhat1 {quantile_hi}% - qhat1"] = df[f"yhat1 {quantile_hi}%"] - self.q_hats[0]
```
ourownstory, line 824: in the case of CQR, I think my comment above also applies - let's simply overwrite `f"yhat1 {quantile_lo}%"` and  `f"yhat1 {quantile_hi}%"`.
IMO, When a user explicitly executes thes conformalize steps, it should be evident enough that the outcome is the conformalized quantile.

neuralprophet/forecaster.py
```diff
@@ -811,6 +813,18 @@ def predict(self, df, decompose=True, raw=False):
             forecast = pd.concat((forecast, fcst), ignore_index=True)
         df = df_utils.return_df_in_original_format(forecast, received_ID_col, received_single_time_series)
         self.predict_steps = self.n_forecasts
+        # Conformal prediction interval with q
+        if self.q_hats:
```
ourownstory, line 817: Can we use a configuration dataclass (or merge it with the QR class) to configure CP?

neuralprophet/forecaster.py
```diff
@@ -811,6 +813,18 @@ def predict(self, df, decompose=True, raw=False):
...
+            if self.conformal_method == "naive":
+                df["yhat1 - qhat1"] = df["yhat1"] - self.q_hats[0]
+                df["yhat1 + qhat1"] = df["yhat1"] + self.q_hats[0]
+            else:  # self.conformal_method == "cqr"
+                quantile_hi = str(max(self.config_train.quantiles) * 100)
```
ourownstory, line 822: This will skip all but the top and bottom quantiles - if a user specifies more than 2, they are not conformalized.
We should either raise an error in this case in the conformalize step (and here), or adopt the conformalize method for more than 2 quantiles. Or am I misreading something here?

neuralprophet/forecaster.py
```diff
@@ -3012,3 +3025,30 @@ def _reshape_raw_predictions_to_forecst_df(self, df, predicted, components):
                         yhat_df = pd.Series(yhat, name=comp).set_axis(df_forecast.index)
                         df_forecast = pd.concat([df_forecast, yhat_df], axis=1, ignore_index=False)
         return df_forecast
+
+    def conformalize(self, df_cal, alpha, method="naive", plotting_backend="default"):
```
ourownstory, line 3029: If a user configures QR, then uses CP, I would expect for them to receive CQR by default. 
It appears to me that the default `method="naive"` will however ignore the QR estimates. 
Maybe we could even remove the `method` arg, as QR + CP = CQR and no-QR + CP = naive CP.
Does that make sense?

neuralprophet/forecaster.py
```diff
@@ -3012,3 +3025,30 @@ def _reshape_raw_predictions_to_forecst_df(self, df, predicted, components):
...
+        """Apply a given conformal prediction technique to get the uncertainty prediction intervals (or q-hats).
+
+        Parameters
+        ----------
+            df_cal : pd.DataFrame
+                calibration dataframe
+            alpha : float
+                user-specified significance level of the prediction interval
```
ourownstory, line 3037: What happens if quantiles in QR were already set? Shouldn't it then be using the same?
We could even make it optional, automatically retrieving the QR quantiles and fail if they are not set, non-symmetric or more than two?

neuralprophet/forecaster.py
```diff
@@ -3012,3 +3025,30 @@ def _reshape_raw_predictions_to_forecst_df(self, df, predicted, components):
...
+            method : str
+                name of conformal prediction technique used
+
+                Options
+                    * (default) ``naive``: Naive or Absolute Residual
+                    * ``cqr``: Conformalized Quantile Regression
+        """
+        df_cal = self.predict(df_cal)
+        if isinstance(plotting_backend, str) and plotting_backend == "default":
```
ourownstory, line 3046: Why do we need to check for `isinstance(plotting_backend, str)`?

neuralprophet/forecaster.py
```diff
@@ -3012,3 +3025,30 @@ def _reshape_raw_predictions_to_forecst_df(self, df, predicted, components):
...
+            plotting_backend = "matplotlib"
+        self.conformal_method = method
```
ourownstory, line 3048: this could also live in  a config

ourownstory (changes requested):
Great work Kevin!

As this exposes a lot of new UI, I focused on critically reviewing those areas.
Thus far, I have only reviewed the `forecaster.py` file. 

Will review rest later, adding a review already so you can move forward with the changes.

neuralprophet/forecaster.py
Kevin-Chen0: CP method is naive, so it does not take into account quantile low and quantile high. That is for CQR. Naive is only for `yhat` +- q, so saying `f"yhat1 {quantile_lo}%"` in incorrect here.

neuralprophet/forecaster.py
Kevin-Chen0: To ensure that `plotting_backend` is not `None`, as it is one of the default values.

neuralprophet/forecaster.py
Kevin-Chen0: I don't agree on this front. The only way it is evident is that is stating `f"yhat1 {quantile_hi}% - qhat1"`.

neuralprophet/forecaster.py
Kevin-Chen0: No, naive is a legitimate conformal prediction method. User should be able to use it regardless of whether there is any QR or not. So my design philosophy is that the user can use whatever CP method regardless of the trained model.

As for what the default method should be, for QR or not QR, that can be debatable. I am leaning towards creating an `"auto"`, to encompass what you said: QR + CP = CQR and no-QR + CP = naive CP.  But that can be added in a new PR.

neuralprophet/forecaster.py
Kevin-Chen0: Not necessarily, you can have `alpha` that is different from QR.

It is possible, by default, to retrieve the QR quantiles, check for symmetry, and automatically alpha, but that can be done as a separate PR.

neuralprophet/forecaster.py
Kevin-Chen0: If the CP is "naive", then NP doesn't use the quantiles at all. If CP is "CQR", then it will take the largest and smallest value in the quantile, regardless of the length of quantiles list.

I can make an assert check that the quantiles list need to have at least 2 values, for high and low, but that can be done as a separate PR.

neuralprophet/forecaster.py
Kevin-Chen0: I'll create a `ConformalPrediction` dataclass in configure.py, which will contain `q_hats`.

neuralprophet/forecaster.py
Kevin-Chen0: I'll create a `ConformalPrediction` dataclass in configure.py, which will contain `q_hats`.

neuralprophet/forecaster.py
Kevin-Chen0: This is resolved, although with the conformal-specific config.

neuralprophet/forecaster.py
Kevin-Chen0: This is done. The dataclass is called `Conformal`.

neuralprophet/conformal_prediction.py
```diff
@@ -0,0 +1,109 @@
...
+        -------
+            list
+                uncertainty prediction intervals (or q-hats)
+
+    """
+    # get non-conformity scores and sort them
+    q_hats = []
+    noncon_scores_list = _get_nonconformity_scores(df_cal, method, quantiles)
+
+    for noncon_scores in noncon_scores_list:
```
ourownstory, line 46: Maybe we can move this part to a separate method/function in a subsequent PR.

neuralprophet/forecaster.py
```diff
@@ -3000,3 +3015,46 @@ def _reshape_raw_predictions_to_forecst_df(self, df, predicted, components):
                         yhat_df = pd.Series(yhat, name=comp).set_axis(df_forecast.index)
                         df_forecast = pd.concat([df_forecast, yhat_df], axis=1, ignore_index=False)
         return df_forecast
+
+    def conformalize(self, df_cal, alpha, method="naive", plotting_backend="default"):
```
ourownstory, line 3019: currently, `predict` implicitly assume that this method has been called beforehand. 
We could:
a) To improve encapsulation we can move the code block from predict here, call this method `conformal_predict` or similar, and have it accept a prediction `df` which is passed on to `predict` internally. 
b) alternatively, this could be split into a three step procedure: `compute_conformalization`, `predict`, `conformalize_prediction`. 

Both would have the advantage of separating regular prediction and conformal prediction.
There may be an even better approach though.

neuralprophet/forecaster.py
```diff
@@ -3000,3 +3015,46 @@ def _reshape_raw_predictions_to_forecst_df(self, df, predicted, components):
...
+                calibration dataframe
+            alpha : float
+                user-specified significance level of the prediction interval
+            method : str
+                name of conformal prediction technique used
+
+                Options
+                    * (default) ``naive``: Naive or Absolute Residual
+                    * ``cqr``: Conformalized Quantile Regression
+            plotting_backend : str
```
ourownstory, line 3034: Maybe we can separate the plotting functionality into a separate method in a subsequent PR.

neuralprophet/forecaster.py
```diff
@@ -3000,3 +3015,46 @@ def _reshape_raw_predictions_to_forecst_df(self, df, predicted, components):
...
+                method=method,
+                q_hats=conformalize(df_cal, alpha, method, self.config_train.quantiles, plotting_backend),
+            )
+        else:
+            self.config_conformal.method = method
+            self.config_conformal.q_hats = conformalize(
+                df_cal, alpha, method, self.config_train.quantiles, plotting_backend
+            )
+
+    # def conformalize_predict(self, df, df_cal, alpha, method="naive"):
```
ourownstory, line 3057: Oh, I see, you already created this method.
Maybe we can make conformalize a util instead of a class method and use this method as the main interface?

neuralprophet/plot_forecast_matplotlib.py
```diff
@@ -131,6 +138,17 @@ def plot(
                         alpha=0.2,
                     )
 
+    # Plot any conformal prediction intervals
```
ourownstory, line 141: for the default plotting method I suggest we only plot one uncertainty type at once, as it may be confusing to most users to see multiple uncertainty types at once.

tests/test_plotting.py
```diff
@@ -539,6 +539,68 @@ def test_plot_uncertainty(plotting_backend):
...
+        m.highlight_nth_step_ahead_of_each_forecast(m.n_forecasts)
+        fig0 = m.plot(forecast, plotting_backend="matplotlib")
+        fig1 = m.plot_components(forecast, plotting_backend="matplotlib")
+        fig2 = m.plot_parameters(plotting_backend="matplotlib")
+        if PLOT:
+            fig0.show()
+            fig1.show()
+            fig2.show()
+    # With auto-regression enabled
+    # TO-DO: Fix Assertion error n_train >= 1
```
ourownstory, line 571: How can we best resolve this issue?

tests/test_uncertainty.py
```diff
@@ -187,3 +186,163 @@ def test_uncertainty_estimation_multiple_quantiles():
...
+        forecast = m.predict(future)
+
+        if PLOT:
+            fig1 = m.plot(forecast)
+            fig2 = m.plot_components(forecast)
+            fig3 = m.plot_parameters()
+            plt.show()
+
+
+"""
```
ourownstory, line 220: what is the reason for these tests being commented out?

neuralprophet/plot_forecast_matplotlib.py
```diff
@@ -422,3 +440,38 @@ def plot_multiforecast_component(
     if multiplicative:
         ax = set_y_as_percent(ax)
     return artists
+
+
+def plot_nonconformity_scores(scores, alpha, q, method):
```
ourownstory, line 445: Very helpful plotting function!

ourownstory (approved):
Good work Kevin!

We can merge this PR, and address the comments in a subsequent PR.

neuralprophet/configure.py
```diff
@@ -426,3 +426,12 @@ def init_holidays(self, df=None):
...
 ConfigCountryHolidays = Holidays
+
+
+@dataclass
+class Conformal:
+    method: str
+    q_hats: list
+
+
+ConfigConformalPrediction = Conformal
```
noxan, line 437: There is no point in having `Conformal` and `ConfigConformalPrediction` right? For other dataclasses we had this separation because they might contain multiple items, yet this is not the case here.

neuralprophet/forecaster.py
```diff
@@ -3000,3 +3015,46 @@ def _reshape_raw_predictions_to_forecst_df(self, df, predicted, components):
...
+
+                Options
+                    * ``None``: No plotting is shown
+                    * ``plotly``: Use the plotly backend for plotting
+                    * ``matplotlib``: Use matplotlib backend for plotting
+                    * ``default`` (default): Use matplotlib backend for plotting
+        """
+        df_cal = self.predict(df_cal)
+        if isinstance(plotting_backend, str) and plotting_backend == "default":
+            plotting_backend = "matplotlib"
```
noxan, line 3045: Why is here another selection of our plotting backend happening? We should definitely have this centralized to make the migration to plotly more easy.

neuralprophet/conformal_prediction.py
```diff
@@ -0,0 +1,109 @@
...
+            user-specified significance level of the prediction interval
+        method : str
+            name of conformal prediction technique used
+
+            Options
+                * (default) ``naive``: Naive or Absolute Residual
+                * ``cqr``: Conformalized Quantile Regression
+
+        quantiles : list
+            list of quantiles for quantile regression uncertainty estimate
```
noxan, line 25: Those conformal prediction rely on the quantile regression as it reuses those parameters or are those separate ones? Just asking as I'm trying to better understand how those to methods are connected (or not)?

neuralprophet/forecaster.py
```diff
@@ -812,6 +816,18 @@ def predict(self, df, decompose=True, raw=False):
...
+            if self.config_conformal.method == "naive":
+                df["yhat1 - qhat1"] = df["yhat1"] - self.config_conformal.q_hats[0]
+                df["yhat1 + qhat1"] = df["yhat1"] + self.config_conformal.q_hats[0]
+            else:  # self.config_conformal.method == "cqr"
+                quantile_hi = str(max(self.config_train.quantiles) * 100)
+                quantile_lo = str(min(self.config_train.quantiles) * 100)
+                df[f"yhat1 {quantile_hi}% - qhat1"] = df[f"yhat1 {quantile_hi}%"] - self.config_conformal.q_hats[0]
+                df[f"yhat1 {quantile_hi}% + qhat1"] = df[f"yhat1 {quantile_hi}%"] + self.config_conformal.q_hats[0]
+                df[f"yhat1 {quantile_lo}% - qhat1"] = df[f"yhat1 {quantile_lo}%"] - self.config_conformal.q_hats[0]
+                df[f"yhat1 {quantile_lo}% + qhat1"] = df[f"yhat1 {quantile_lo}%"] + self.config_conformal.q_hats[0]
```
noxan, line 830: I'm a bit confused with all the extra output values. What benefit do they bring me as a user and how would I interpret them? Maybe conformal prediction is an advanced feature overall, yet it would be great to have some guide or instructions on how to make use of this method or at least to better understand what I'm missing out on 😅

neuralprophet/forecaster.py
noxan: I would definitely love to see a more unified approach to our different uncertainty modelling approaches - it's getting rather confusing to me and also how different their APIs are or rather my uncertainty on how to use them / combine them.

neuralprophet/configure.py
Kevin-Chen0: Not necessarily for `ConfigConformalPrediction`, but @ourownstory wants me to put `q_hats` and `method` into a conformal/uncertainty config dataclass instead of being a primary class variable for `NeuralProphet`. So I did.

neuralprophet/forecaster.py
Kevin-Chen0: It's needed to order to print out the **One-Sided Interval Width with q** plot when user is running `.conformalize()`. This plot is very helpful in visualizing the `q` for given `alpha`, where `q` is the basis of the prediction interval. See the [uncertainty_conformal_prediction.ipynb](https://github.com/ourownstory/neural_prophet/blob/main/tutorials/feature-use/).

neuralprophet/conformal_prediction.py
Kevin-Chen0: Right now there are two available conformal prediction methods: **naive** and **CQR**. Naive does conformal prediction independently from the QR while CQR applies conformal prediction on the QR itself rather than the point prediction.

neuralprophet/conformal_prediction.py
```diff
@@ -0,0 +1,109 @@
...
+                max(row[f"yhat1 {quantile_lo}%"] - row["y"], row["y"] - row[f"yhat1 {quantile_hi}%"]),
+                0 if row[f"yhat1 {quantile_lo}%"] - row["y"] > row["y"] - row[f"yhat1 {quantile_hi}%"] else 1,
+            ]
+        )
+        scores_df = df.apply(cqr_scoring_func, axis=1, result_type="expand")
+        scores_df.columns = ["scores", "arg"]
+        scores_list = [scores_df["scores"].values]
+    else:  # method == "naive"
+        # Naive nonconformity scoring function
+        scores_list = [abs(df["y"] - df["yhat1"]).values]
```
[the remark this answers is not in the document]
Kevin-Chen0: Yes, it is hardcoded so it only does one-step ahead. I agree we can create a subsequent PR to enable conformal prediction for multiple steps.

neuralprophet/conformal_prediction.py
```diff
@@ -0,0 +1,109 @@
...
+        quantiles : list
+            list of quantiles for quantile regression uncertainty estimate
+
+        Returns
+        -------
+            list
+                nonconformity scores from the calibration datapoints
+
+    """
+    quantile_hi = None
```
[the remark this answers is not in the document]
Kevin-Chen0: I don't follow. CQR only looks at the lowest and highest quantiles regardless of the number of quantiles in-between (if any). Maybe can extend for more advanced versions of conformal prediction, although I don't know of any that need beyond high and low quantiles.

tests/test_uncertainty.py
Kevin-Chen0: The commented tests are moved to `test_plotting.py`, but I forgot the delete the comment `test_uncertainty.py`. I can remove that comments in the next PR.

tests/test_plotting.py
Kevin-Chen0: Will need to dig into it.

neuralprophet/forecaster.py
```diff
@@ -812,6 +816,18 @@ def predict(self, df, decompose=True, raw=False):
             forecast = pd.concat((forecast, fcst), ignore_index=True)
         df = df_utils.return_df_in_original_format(forecast, received_ID_col, received_single_time_series)
         self.predict_steps = self.n_forecasts
+        # Conformal prediction interval with q
```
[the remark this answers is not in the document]
Kevin-Chen0: No, I have it when the `.predict()` outputs the forecast_df, that df also contains the forecasted value plus/min the conformal prediction interval. For example:

yhat1 - qhat1 | yhat1 + qhat1
-- | --

Just like what's been with QR, given forecast columns like: 

yhat1 5.0% | yhat1 95.0%
-- | --

tests/test_uncertainty.py
Kevin-Chen0: The commented tests have been removed with this PR #1032.

neuralprophet/forecaster.py
Kevin-Chen0: Possible, but I intend to give the user the to run either the `.conformalize()` and `.predict()` methods separately or run as one with `.conformalize_prediction()`, just like for scikit's `.fit_transform()`. However, the reason why I haven't yet is because the `.conformalize_prediction()` will need to input two datasets, on the calibration set and another test set. Maybe that is fine because b/c the `.train()` gives the optionality of adding a validation set alongside the train set.

neuralprophet/forecaster.py
Kevin-Chen0: Please see PR #1044. This implements what you have for **a)**. There is a `conformal_predict()` method that combines the `conformalize()` and `predict()`, and will need to input most the calibration and test sets for split CP. With this implemented, `conformalize()` will then be removed.

Kevin-Chen0: Also, for future bootstrapped (e.g. Jackknife+ and CV+ based) CP methods, calibration set will not be needed, which is why `calibration_df` is set to `None` as default in `conformal_predict()`.

PR #819, opened 2022-10-14
[bugfix] Fixed wrong return values for get_last_forecast() and plot_last_forecast()

### LINKED ISSUE #815: [bug] Check correct return values for get_latest_forecast and plot_last_forecast
should return entire n_forecast predictions at the last forecast origin. (and potentially include a few older forecasts)

judussoari:
Closes #815 
The return values of `get_last_forecast()` (renamed from get_latest_forecast() for consistency) and `plot_last_forecast()` were wrong, which potentially causes undesirable consequences. 

*The term "last forecast" corresponds to the row of yhats of the last timestamp

*e.g. `yhat1` is the forecast made on the very last timestamp

*`yhat2` is the forecast made on the second to last timestamp


ourownstory:
closing in favor of naming fix

PR #829, opened 2022-10-15
Rename "yhat" to "origin" in *_latest_forecast() to prevent confusion

judussoari:
As of now, both the regular forecast and the latest forecast df columns were named "yhat*". However, those columns are composed differently and must therefore not be named the same 

Additionally, `plot_last_forecast()` has been renamed to `plot_latest_forecast()` for consistency

When calling `plot_latest_forecast()`, the quantiles (if specified) for the very last forecast (called `origin-0`) are plotted in both plotly and matplotlib


neuralprophet/plot_forecast.py
```diff
@@ -81,9 +81,14 @@ def plot(
     else:
         fig = ax.get_figure()
     ds = fcst["ds"].dt.to_pydatetime()
-    yhat_col_names = [col_name for col_name in fcst.columns if "yhat" in col_name]
-
-    if highlight_forecast is None or line_per_origin:
+    colname = "yhat"
+    latest = 1
```
ourownstory, line 85: rename step

ourownstory (changes requested):
please also update the plotly version

Issue #803, opened 2022-10-11
Fix pandas future warnings in df_utils.py


noxan:
```
df_utils.py:193: FutureWarning: In a future version, `df.iloc[:, i] = newvals` will attempt to set the values inplace instead of always setting a new array. To retain the old behavior, use either `df[df.columns[i]] = newvals` or, if columns are non-unique, `df.isetitem(i, newvals)`
    df.loc[:, "ds"] = pd.to_datetime(df.loc[:, "ds"])
```

PR #839, opened 2022-10-17
[refactor] Adress warnings

karl-richter:


noxan:
@karl-richter does this pull request also fix #803?

neuralprophet/df_utils.py
```diff
@@ -193,8 +193,8 @@ def data_params_definition(df, normalize, config_covariates=None, config_regress
 
     data_params = OrderedDict({})
     if df["ds"].dtype == np.int64:
-        df.loc[:, "ds"] = df.loc[:, "ds"].astype(str)
-    df.loc[:, "ds"] = pd.to_datetime(df.loc[:, "ds"])
+        df = df.assign(ds=df.loc[:, "ds"].astype(str))
+    df = df.assign(ds=pd.to_datetime(df.loc[:, "ds"]))
```
karl-richter, line 197: @ourownstory what do you think about the `assign` function? According to pandas this seems to be the "official way" to assign values to a column. However, I think we slightly compromise on readability compared to `df["ds"]=...`. Should we remain with `assign()` or use the plain version?

karl-richter (commented):
Added question for reviewers

neuralprophet/df_utils.py
karl-richter, line 197: @ourownstory what do you think about the `assign` function? According to pandas this seems to be the "official way" to assign values to a column. However, I think we slightly compromise on readability compared to `df["ds"]=...`. Should we remain with `assign()` or use the plain version?

ourownstory (approved):
LGTM

PR #842, opened 2022-10-18
fix black warning (remove additional newline)

noxan:


ourownstory:
Shouldn't this have been caught by #878 when I ran black over all files?

PR #853, opened 2022-10-18
[feature] Support setting custom quantile in plot_parameters()

LeonieFreisinger:
Modify the plot_parameters() function to allow the user to plot a chosen quantile instead of the median.  

Therefore, the matplotlib and plotly version will be modified. The quantile will be plotted with absolute value.


ourownstory:
@LeonieFreisinger To simplify things, just change all panels to the specific set quantile (instead of the (default) median). 
So, if people call plot_parameters with no args, it plots the parameters of the median. 
If they set the quantile, it plots (only) the parameters of the set quantile.

LeonieFreisinger:
@ourownstory custom quantile for plotting parameters and respective tests are implemented. Would be great, if you could review it. 
Thanks!

neuralprophet/forecaster.py
```diff
@@ -1846,6 +1854,13 @@ def plot_parameters(
                 raise Exception(
                     "df_name parameter is required for multiple time series and local modeling of at least one component."
                 )
+        if quantile is not None:
+            # ValueError if model was not trained or predicted with selected quantile for plotting
+            if not (0 < quantile < 1):
+                raise ValueError("The quantile selected needs to be a float in-between (0,1)")
+            # ValueError if selected quantile is out of range
+            if quantile not in self.config_train.quantiles:
+                raise ValueError("Selected quantile is not specified in the NeuralProphet object.")
```
LeonieFreisinger, line 1863: model configuration

ourownstory (approved):
LGTM

PR #855, opened 2022-10-18
[feature] shared weight nn_net

LeonieFreisinger:
Referencing previous PR #616 

**Problem**
Currently, the weights of the separate implementations of the time series components (AR Net, Time Net, Covariates) are updated independently. Hence, they are not updated in dependence on each other.

**Solution** 
Implement a shared weight update and allow the user to choose between shared and independent weights.


LeonieFreisinger:
PR will be further processed by other contirbutor.

Issue #834, opened 2022-10-17
Add Github problem matcher to Github test and black actions


noxan:
https://github.com/liskin/gh-problem-matcher-wrap

PR #859, opened 2022-10-19
834 add GitHub problem matcher to GitHub test and black actions

noxan:
#834


ourownstory (approved):
Sweet, looking forward to this!

Seems like it is just for pytests set up - not black?

PR #865, opened 2022-10-19
Allow component plot with locally-normalized global df

### LINKED ISSUE #807: [enhancement] allow residuals plotting with local normalization
currently disabled in `plot_components`, but should not be hard to allow

judussoari:
Closes #807 
`plot_components()` threw an Error whenever a global modal has been fitted with local normalization

The desired ID of the global df shall now be passed as an argument, e.g. `plot_components(forecast, df_name='df1')`.
If `df_name` is not specified (i.e. no df with multiple IDs), the default name `__df__` is used.

### DIFF
--- a/neuralprophet/forecaster.py
+++ b/neuralprophet/forecaster.py
@@ -1698,7 +1698,7 @@ class NeuralProphet:
             )
 
     def plot_components(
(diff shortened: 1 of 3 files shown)


ourownstory (approved):
LGTM

PR #867, opened 2022-10-20
[feature] NN shared weights

karl-richter:
> Builds on #837, closes #616 

**Changes:**
- Use a single network for both autoregression and covariates


karl-richter:
This pr is outdated. If the feature is desired in the future, a new branch should be created.

PR #869, opened 2022-10-20
868 bug runtime exception during collection when declaring configlaggedregressors

### LINKED ISSUE #868: [bug] Runtime exception during collection when declaring ConfigLaggedRegressors
Refers to PR #805 

Error during collection occurs in `configure.py` because OrderedDict is subscripted/ not declared right 

`ConfigLaggedRegressors = OrderedDict[str, LaggedRegressor]`

Pylance: `Subscript for class "OrderedDict" will generate runtime exception; enclose type annotation in quotes`

Quick fix suggestion: put in quotes `ConfigLaggedRegressors = "OrderedDict[str, LaggedRegressor]"`

noxan:
Two problems are fixed with this PR:
- Fix for #868: OrderedDict does behave differently as typing across python version apparently, super weird (see other [issues](https://github.com/python/mypy/issues/6904))
- Fixes missing `typing.Literal` issue with python version 3.7

### DIFF
--- a/neuralprophet/configure.py
+++ b/neuralprophet/configure.py
@@ -4,6 +4,7 @@ import inspect
 import logging
 import math
 import types
+import typing
 from collections import OrderedDict
 from dataclasses import dataclass, field
 from typing import Callable, Optional, Union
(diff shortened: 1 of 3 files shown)


ourownstory (approved):
Thank you for the quick fix!

btw, for the new requirement, should we only require it for certain pyhton versions?

noxan:
> Thank you for the quick fix!
> 
> btw, for the new requirement, should we only require it for certain pyhton versions?

Great idea, added another PR https://github.com/ourownstory/neural_prophet/pull/873

PR #870, opened 2022-10-20
fix command "run" not found in github website deploy action

noxan:


ourownstory (commented):
Seems like this was already merged in #871?

ourownstory:
Seems like this was already merged in #871?

PR #873, opened 2022-10-20
only require python typing extensions for python < 3.8

noxan:
typing.Literal exists with python 3.8, not prior
https://docs.python.org/3.8/library/typing.html#typing.Literal


ourownstory (approved):
LGTM

PR #881, opened 2022-10-20
[dev-ops] Pre commit isort new files

ourownstory:


noxan:
@ourownstory Looks all good to me. I've merged the two loops on python files, makes it more simple.

Also for me the commit message had been "Sorted by isort", even when files were formatted with black - is there some way to adjust that message or make it differ?

karl-richter (approved):

PR #883, opened 2022-10-21
[model] Deep model parameter interpretation

karl-richter:
**Status quo**
When the NN uses hidden layers, only the model weights of the first layer are interpreted when using the `plot_parameters()` function. This could be misleading to a user who wants to interprete the model.

> Note: in some places this lead to wrong results in the plots, eg. when a user set `highlight_nth_forecast` in `plot_parameters()`, the weights of the lags w.r.t the n'th hidden layer is returned (since the weights are `input x hidden1`). If the shape of the hidden layer does not match the shape of the output, this also leads to runtime issues.

**Change**
Instead of using the weights of the first layer, we use a model attribution method to calculate the attributions of the lags w.r.t. each forecast. We use pytorch's captum library for saliency calculation.

### DIFF
--- a/neuralprophet/time_net.py
+++ b/neuralprophet/time_net.py
@@ -5,6 +5,7 @@ import torch
 import torch.nn as nn
 import logging
 from neuralprophet import configure
+from neuralprophet.utils_torch import interprete_model
 from neuralprophet.utils import (
     config_season_to_model_dims,
     config_regressors_to_model_dims,
@@ -402,6 +403,26 @@ class TimeNet(nn.Module):
             regressor_params = self.regressor_params["multiplicative"]
 
         return regressor_params[:, index : (index + 1)]
+
+    def get_model_input_attributions(self, net: str = "ar_net", forward_func: str = "auto_regression"):
+        """
+        Returns model input attributions for a given network and forward function.
+
+        Parameters
+        ----------
+            net : str
+                Name of the network for which input attributions are to be computed.
+
+            forward_func : str
+                Name of the forward function for which input attributions are to be computed.
+
+        Returns
+        -------
+            torch.Tensor
+                Input attributions for the given network and forward function.
+        """
+        attributions = interprete_model(self, net, forward_func)
+        return attributions
 
     def _compute_quantile_forecasts_from_diffs(self, diffs, predict_mode=False):
         """
--- a/neuralprophet/utils_torch.py
+++ b/neuralprophet/utils_torch.py
@@ -122,10 +122,13 @@ def interprete_model(target_model, net, forward_func):
     model = InterpretableModel(target_model, forward_func)
     saliency = Saliency(model)
 
+    # Number of quantiles
+    num_quantiles = len(target_model.quantiles)
     # Number of input features to the net (aka n_lags)
     in_features = getattr(target_model, net)[0].in_features
     # Number of output features from the net (aka n_forecasts)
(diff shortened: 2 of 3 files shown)


neuralprophet/utils_torch.py
```diff
@@ -96,3 +97,50 @@ def create_optimizer_from_config(optimizer_name, model_parameters, lr):
...
+            Name of the forward function for which input attributions are to be computed.
+
+    Returns
+    -------
+        torch.Tensor
+            Input attributions for the given network and forward function.
+    """
+    # Load the respective forward function from the model and init model interpreter
+    forward = getattr(target_model, forward_func)
+    saliency = Saliency(forward_func=forward)
```
karl-richter, line 121: fyi - I got rid of the "InterpretableModel" here and simply pass the forward function to the attribution method

karl-richter (commented):
fyi

neuralprophet/time_net.py
```diff
@@ -7,6 +7,7 @@
 import torch.nn as nn
 
 from neuralprophet import configure
+from neuralprophet.utils_torch import interprete_model
```
ourownstory, line 10: can remove

neuralprophet/utils_torch.py
```diff
@@ -97,3 +98,50 @@ def create_optimizer_from_config(optimizer_name, model_parameters, lr):
...
+
+    # Number of quantiles
+    num_quantiles = len(target_model.quantiles)
+    # Number of input features to the net (aka n_lags)
+    in_features = getattr(target_model, net)[0].in_features
```
ourownstory, line 127: `num_`...

ourownstory (dismissed):
great looking!
final task: find and update notebooks

Kevin-Chen0:
@karl-richter Can you resolve merge conflict and fix flake8? Thx

noxan:
@ourownstory You asked about some notebook updates - is this still open or can we merge this PR?

neuralprophet/utils_torch.py
```diff
@@ -49,3 +50,49 @@ def create_optimizer_from_config(optimizer_name, optimizer_args):
...
         raise ValueError("The provided optimizer is not supported.")
     return optimizer, optimizer_args
+
+
+def interprete_model(target_model, net, forward_func):
+    """
+    Returns model input attributions for a given network and forward function.
+
+    Parameters
+    ----------
```
noxan, line 60: `target_model` is not defined as parameter in docstring

neuralprophet/utils_torch.py
```diff
@@ -49,3 +50,49 @@ def create_optimizer_from_config(optimizer_name, optimizer_args):
     else:
         raise ValueError("The provided optimizer is not supported.")
     return optimizer, optimizer_args
+
+
+def interprete_model(target_model, net, forward_func):
```
noxan, line 55: would be great to have typings ;)

noxan (approved):
LGTM - very clean and structured code, was good to review even without knowing all the specifics :)

Two minor remarks, see comments.

neuralprophet/utils_torch.py
karl-richter: Done

neuralprophet/utils_torch.py
karl-richter: Done

Issue #875, opened 2022-10-20
Allow plotting of only specific modules when calling plot_parameters and plot_components


ourownstory:
add optional argument (list of component names) that specifies which panels to plot.

PR #884, opened 2022-10-21
[feature] Support plotting of only specific panels in plot_parameters and plot_components

LeonieFreisinger:
Referencing issue #875 
**Problem**
The current implementation of `plot_parameters() `and `plot_components()` by default plots all parameters or components that are specified by the user in the model configuration. However, the `plot_parameters()` and `plot_components()` should allow for plotting only specific panels. 

**Solution**
Add an optional argument (list of component names) in `plot_parameters()` and `plot_components()` to specify which panel to plot.

 **Key changes**
- consolidate code from the plotting-backend specific functions in the top-level `plot_parameters()` and `plot_components()` to reduce double maintenance effort.
- introduce `get_validated_components()` to validate the selected components by the user.
- introduce `check_if_configured() `to check if input components are configured.
- add `forecast_in_focus` as new input to `plot()` to be consistent with `plot_parameters()` and `plot_components()`. `forecast_in_focus` can be used to overwrite `self.highlight_forecast_step_n`

remark: 
- the user can only choose all events at once, i.e. added events and country holidays.

**Review Checklist**

- [x]  I have performed a self-review of my own code.
- [x] My code follows the style guidelines described under CONTRIBUTING.md. 
- [x] I have added pytests to check whether my feature / fix works
- [x] All gh actions pass or the failure is not related to my code.


neuralprophet/forecaster.py
```diff
@@ -1736,6 +1742,21 @@ def plot_components(
                 * ``matplotlib``: use matplotlib for plotting
                 * (default) ``default``: use the global default for plotting
 
+            plotting_panels: str or list, optional
```
ourownstory, line 1745: can call 'components'

neuralprophet/forecaster.py
```diff
@@ -1736,6 +1742,21 @@ def plot_components(
...
+                ----
+                * (default)``None``:  All components the user set in the model configuration are plotted.
+                * ``trend``
+                * ``seasonality``
+                * ``ar``
+                * ``lagged_regressors``
+                * ``future_regressors``
+                * ``events``
+                * ``residuals``
+                * ``quantiles``
```
ourownstory, line 1758: lets call this uncertainty

neuralprophet/forecaster.py
```diff
@@ -1736,6 +1743,23 @@ def plot_components(
...
 
+            components: str or list, optional
+                name or list of names of components to plot
+
+                Options
+                ----
+                * (default)``None``:  All components the user set in the model configuration are plotted.
+                * ``trend``
+                * ``seasonality``
+                * ``auto-regression``
```
LeonieFreisinger, line 1754: autoregression

LeonieFreisinger:
@alfonsogarciadecorral and @Kevin-Chen0 it would be great if one of you can give my code a quick review
CC: @ourownstory 
Thanks a lot!

Currently, the model/metrics and test/docs are still failing due to an overall issue. However, @karl-richter is already aware of the model/metrics issue and @noxan is aware of the tests/docs issue.

neuralprophet/forecaster.py
```diff
@@ -1878,6 +1957,35 @@ def plot_parameters(
...
+                    "trend",
+                    "seasonality",
+                    "autoregression",
+                    "lagged_regressors",
+                    "future_regressors",
+                    "country_holidays",
+                    "events",
+                ]:
+                    raise ValueError(" Selected component(s) for plotting are mis-spelled.")
+
```
noxan, line 1988: Can't this validation be extracted to some helper function as we obviously have it twice (very prone to forgetting changes in the other copy)?

neuralprophet/forecaster.py
```diff
@@ -1772,6 +1796,44 @@ def plot_components(
...
+                        and ("events_additive" not in fcst.columns and "events_multiplicative" not in fcst.columns)
+                    )
+                    or (
+                        "future_regressors" in comp
+                        and (
+                            "future_regressors_additive" not in fcst.columns
+                            and "future_regressors_multiplicative" not in fcst.columns
+                        )
+                    )
+                    or ("uncertainty" in comp and not len(self.model.quantiles) > 1)
```
noxan, line 1822: Idea: As the checks whether a model component is actually enabled is implemented several times and might be useful for other functionalities (e.g. configuration management) of our framework, it might be a good idea to extract the implementation and unify those checks (I stumbled across this while rewriting parts of the configurations)

noxan (changes requested):
Looks like a great set of changes and exactly what I want to use for our new tutorials. Just stumbled across the extra `s` in the github flow (that was not exactly the failing test I meant). Added a few other ideas for potential improvements (possibly outside this PR). Happy to discuss the details.

.github/workflows/website.yml
```diff
@@ -38,7 +38,7 @@ jobs:
       - name: Build with Sphinx
         run: sphinx-build docs/source _site
       - name: Upload artifact
-        uses: actions/upload-pages-artifact@v1
+        uses: actions/upload-pages-artifact@v1s
```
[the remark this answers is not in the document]
LeonieFreisinger: Thanks @noxan !

LeonieFreisinger:
> PR). Happy to discuss the details. Gr

Amazing! Thanks @noxan, I will rework the code according to your ideas.

neuralprophet/forecaster.py
LeonieFreisinger: @noxan thanks for the input! I think this is a great idea. Let´s sync on this.

alfonsogarciadecorral:
Hi @LeonieFreisinger 

Sorry for not reviewing before! I just had a look at the code and 've been playing with the new functionality. 

It works really nicely. So nothing to add from me!

LeonieFreisinger:
@alfonsogarciadecorral thanks a lot!
@noxan I implemented the changes we discussed. Would be great if you could have a final review. Thank you.

neuralprophet/forecaster.py
```diff
@@ -1877,6 +2379,28 @@ def plot_parameters(
             # ValueError if selected quantile is out of range
             if quantile not in self.config_train.quantiles:
                 raise ValueError("Selected quantile is not specified in the model configuration.")
+        else:
+            # plot parameters for median quantile if not specified
+            quantile = self.config_train.quantiles[0]
+
+        # Validate components to be plotted
+        valid_set = [
```
noxan, line 2387: It's a bit hard to figure out the difference between components and parameters, so I'd propose be better name them more explicitly.

```suggestion
        valid_components_set = [
```

neuralprophet/forecaster.py
```diff
@@ -1877,6 +2379,28 @@ def plot_parameters(
...
+            "autoregression",
+            "lagged_regressors",
+            "events",
+            "future_regressors",
+        ]
+        valid_components = self.get_validated_components(
+            components=components,
+            df_name=df_name,
+            forecast_in_focus=forecast_in_focus,
+            valid_set=valid_set,
```
noxan, line 2400:
```suggestion
            valid_set=valid_components_set,
```

neuralprophet/plot_forecast.py
```diff
@@ -193,140 +190,11 @@ def plot_components(
...
-    elif len(m.model.quantiles) > 1 and forecast_in_focus is not None:
-        for i in range(1, len(m.model.quantiles)):
-            components.append(
-                {
-                    "plot_name": "Uncertainty",
-                    "comp_name": f"yhat{forecast_in_focus} {round(m.model.quantiles[i] * 100, 1)}%",
-                    "fill": True,
-                }
-            )
+    plot_components = components["components"]
```
noxan, line 193: Was a bit hard for me to understand what `components["components"]` is here. Maybe we find some more explicit name for the dictionary itself and for the "components to plot" field within that dictionary. e.g. 
```
components_to_plot = plot_configuration["component_list"]
```
or something similar 🤔

.github/workflows/website.yml
noxan, line 41:
```suggestion
        uses: actions/upload-pages-artifact@v1
```

noxan (changes requested):
Was a way bigger refactor than I intended, yet happy to see we have less code now overall.

The original additional `s` in the website workflow is still there. Also added a few other comments, which are rather of cosmetic nature - if the website workflow works again I'm happy - many thanks! :)

neuralprophet/forecaster.py
```diff
@@ -1711,8 +1735,412 @@ def plot_latest_forecast(
...
+            error_flag : bool
+                Activate to raise a ValueError if component has not been configured
+
+        Returns
+        -------
+            components
+                list of components only including the components set in the model configuration
+        """
+        # Error if selected plotting components are not set in the model configuration by the user
+        for comp in reversed(components):
```
noxan, line 1766: If I understood the implementation right, this seems a bit inefficient - we validate all components for each component. Wouldn't it be better to only validate each component once (independent of the others)?

noxan: Also why was this moved into `forecaster.py`? To me this is all purely plotting related and would better remain in `plot_forecast.py`

neuralprophet/forecaster.py
```diff
@@ -1711,8 +1735,412 @@ def plot_latest_forecast(
                 line_per_origin=True,
             )
 
+    def check_if_configured(self, components, error_flag=False):
```
noxan, line 1738: Why was this function moved into `forecaster.py`? To me this is all purely plotting related and would better remain in `plot_forecast.py`

Or if we want to turn this into a generic implementation for checking whether a component is configured, it would be better to allow validation of each component individually and separate the configuration check from the plotting parameter generation - guess all only relevant for another PR, I was surprised by the amount of change here.

neuralprophet/forecaster.py
```diff
@@ -1711,8 +1735,412 @@ def plot_latest_forecast(
...
+        -------
+            components
+                list of components only including the components set in the model configuration
+        """
+        # Error if selected plotting components are not set in the model configuration by the user
+        for comp in reversed(components):
+            if (
+                ("trend_rate_change" in comp and self.model.config_trend.changepoints is None)
+                or ("seasonality" in comp and self.config_season is None)
+                or ("autoregression" in comp and not self.n_lags > 0)
```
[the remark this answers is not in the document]
LeonieFreisinger: > Was a way bigger refactor than I intended, yet happy to see we have less code now overall.
> 
> The original additional `s` in the website workflow is still there. Also added a few other comments, which are rather of cosmetic nature - if the website workflow works again I'm happy - many thanks! :)

Hey @noxan thanks for taking the time to review again!! :) Yes, was also a little surprised that it turned out to be a larger refactor, but happy to have reduced some code lines. sorry about the s in the website workflow! Will keep an eye on it.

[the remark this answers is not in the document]
LeonieFreisinger: @noxan regarding the `n_lags`: in forecaster.py in line 387 `self.n_lags = self.config_ar.n_lags`. I assumed it makes no difference. Happy to change it though, if you explain the intention behind it.

neuralprophet/forecaster.py
```diff
@@ -1711,8 +1735,412 @@ def plot_latest_forecast(
...
+                "lagged_scalar_regressors": lagged_scalar_regressors,
+                "overwriting_unknown_data_normalization": overwriting_unknown_data_normalization,
+                "df_name": df_name,
+            }
+        elif validator == "plot_components":
+            output_dict = {
+                "components": plot_components,
+            }
+        return output_dict
+
```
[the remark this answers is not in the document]
LeonieFreisinger: good idea!

neuralprophet/forecaster.py
LeonieFreisinger: Definitely true ^^ will change

neuralprophet/forecaster.py
LeonieFreisinger: @noxan ok, I see. Would you recommend moving them to `plot_forecaster.py`?  I agree, with you to keep the plotting-related functions separated in order to not overload `forecaster.py`. Since the `check_if_configured()` is used for components and parameters, and the respective functions are in different .py files, is the most proper way to define the function in the `plot_forecaster.py`?

neuralprophet/forecaster.py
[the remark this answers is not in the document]
noxan: Just thought it would be more consistent to also check `self.config_ar` (we checked the config for every other component, so why not also for this one)

neuralprophet/forecaster.py
noxan: I think it might be best to create a new file for code which is shared across our two plotting libraries. We could name it `plot_forecaster_shared.py` or something 🤔 

Reasoning: I would prefer to put it  in a file named `plot_forecaster.py`. But currently `plot_forecaster.py` is implicitly the `matplotlib` implementation, in contrast to the explicit `plotly` implementation which has the proper prefix in the filename. So this does not make much sense.
Future vision: we rename the current `plot_forecaster.py` to `plot_forecaster_matplotlib.py` and name the file for the shared code properly `plot_forecaster.py`

ourownstory (approved):
Great work Leonie!
This should make future work on plotting easier and allow for more flexibility. Wonderful!

Maybe we could move the validation utilities to a separate file, for better organization, but we can revisit that later.

Kevin-Chen0:
@LeonieFreisinger can you resolve the merge conflicts? After that, @ourownstory @noxan, is this ready to be approved? Thanks.

noxan (approved):
LGTM, thanks for bearing with all the iterations @LeonieFreisinger, many thanks :)

PR #886, opened 2022-10-21
Flake8 fixes

noxan:
Cherry picking commits from #833


ourownstory (approved):
LGTM. 
Seems like there are many less changes than in #833 - is that on purpose?

noxan:
> LGTM. Seems like there are many less changes than in #833 - is that on purpose?

Yes, thanks for paying close attention. 833 contained the flake8 fixes, new github actions and isort fixes, which are now 3 dedicated PRs. maybe some minor changes are missing, in case flake8 will fail and we know :)

PR #887, opened 2022-10-21
825 GitHub actions for isort and flake8 linters

noxan:
Replacement for #833 containing the original flake8 and isort github actions


ourownstory (approved):
LGTM - I wonder if it may be better to the most recent python version for this?

noxan:
> LGTM - I wonder if it may be better to the most recent python version for this?

Good point, adjusted and merged.

PR #888, opened 2022-10-21
[feature] Plot components of averaged df (and quants) if multiple IDs, but no df_name specified in plot_components()

### LINKED ISSUE #866: [feature] plot components of average and std dev of all dfs in global df, if df_name not specified
- [ ] Identify when there are multiple IDs in df
- [ ] calculate average and std
- [ ] see if std dev or max/min makes more sense to plot (as shade)
- [ ] Include in `plot_components()` for matplotlib and plotly

judussoari:
Closes #866 

To Do:
- [x] Add mean plot of components across all dfs
- [x] Add Standard Deviation to plot of averaged df components (as a shade in both directions)
- [x] Migrate the code to plot_parameters() instead of plot_cmponents, makes more sense

Before: We raised an Exception whenever multiple time series were present in the input df and `plot_paramenters()` was called without specifying the time series to be plotted. 
Now: plots mean and quants of all time series when time series is not specified in `df_name`. 

Sceenshot 1: model parameters when `df_name` is specified. 
Sceenshot 2: model parameters when `df_name` is **not** specified. 
Bildschirmfoto 2022-11-21 um 23 48 06
Bildschirmfoto 2022-11-21 um 23 47 55


alfonsogarciadecorral:
HI @judussoari ! Really nice work, this is super useful!
I have a few comments, I will put them below.

Computation time. plotting parameters is talking long. But that is because of numpy on the .groupby().apply()

If you change:
```
            df_trend_std = df_trend.groupby("ds")[["trend"]].apply(lambda x: np.std(x))
            df_trend = df_trend.groupby("ds")[["trend"]].apply(lambda x: np.mean(x))
```
by:
```
            df_trend_std = df_trend.groupby("ds")[["trend"]].apply(lambda x: x.std())
            df_trend = df_trend.groupby("ds")[["trend"]].apply(lambda x: x.mean())
```

it is much faster.

alfonsogarciadecorral:
I think it would be more intuitive to work with quantiles? 0.2,0.5 and 0.8 for example.

Trends won't follow normal distributions across the panel and that might cause issues/misleading results. 

For example with mean -std we can see negative values of trend. I attach an example using ercot data (plotly is with quantiles and matplotlib with the mean and stds)

Let me know what you think!
Screenshot 2022-10-27 at 10 49 27
Screenshot 2022-10-27 at 10 49 39

Code used:
```
            df_trend_q80 = df_trend.groupby("ds")[["trend"]].apply(lambda x: x.quantile(0.8))
            df_trend_q20 = df_trend.groupby("ds")[["trend"]].apply(lambda x: x.quantile(0.2))
            df_trend = df_trend.groupby("ds")[["trend"]].apply(lambda x: x.quantile(0.5))
```

alfonsogarciadecorral:
It could be useful to add add a legend on the matplotlib version, explaining what is upper bound, lower bound and the "main line".

judussoari:
@alfonsogarciadecorral thanks for your super helpful suggestions! Implemented them with one exception: I am not plotting the median quantile, but stick with the mean across all IDs instead. Otherwise the plot does not look intuitive at all, unfortunately. Also got ok of @ourownstory. 
Ready to merge!

Kevin-Chen0:
@judussoari can you resolve the merge conflicts? Thanks.

Kevin-Chen0 (approved):
LGTM

PR #894, opened 2022-10-24
add matrix for python versions 3.7 and 3.10 in tests

### LINKED ISSUE #874: Add python 3.7 to tests in github action
- add python 3.7
- but run it only on ubuntu

noxan:
Close #874


noxan:
https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python

ourownstory (approved):
So we will just run it over all OS? LGTM, as long as it does not take too long. :)

PR #897, opened 2022-10-24
778 update docstrings to match typing of variables

### LINKED ISSUE #778: Update docstrings to match typing of variables
Seems like some docstrings were not updated to our switch to using `@dataclass class` instead of `OrderedDict`.
Could you update all the docstrings (seems like primarily in time_net.py) that still mention `OrderedDict` in a separate PR?

_Originally posted by @ourownstory in https://github.com/ourownstory/neural_prophet/pull/714#discussion_r971077972_

noxan:
Close #778

### DIFF
--- a/neuralprophet/configure.py
+++ b/neuralprophet/configure.py
@@ -394,6 +394,9 @@ class Event:
     mode: str
 
 
+ConfigEvents = OrderedDictType[str, Event]
+
+
 @dataclass
 class Holidays:
     country: str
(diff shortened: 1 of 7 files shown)


ourownstory (changes requested):
Looking at this, I think we could be more consistent with the other changes you are making (e.g. `ConfigFutureRegressors`).

Let's call `config_regressors` -> `config_future_regressors` 
Let's just be careful to avoid user-facing api changes for now.

noxan:
Yes, absolutely agree on the goal to rename all variables and make sure we have typings everywhere. Let's make those changes gradually (keep PRs in manageable chunks) and stick to the scope of the issues (this one being about docstring types) - my 2 cents, happy to discuss other approaches too :)

PR #898, opened 2022-10-24
[fix] added kaleido dev requirement

karl-richter:
As noted by @alfonsogarciadecorral, when manually running the benchmark tests locally, they currently fail since the kaleido package is not installed that is used to export the plotly figures. This pr adds the package to the dev requirements.


alfonsogarciadecorral (approved):
Working!

PR #900, opened 2022-10-25
fix docs requirements path and filename

noxan:
@ourownstory thanks for finding my mistake, should have been part of the other PR already, my bad


ourownstory (approved):
wonderful!

PR #903, opened 2022-10-25
[major] lagged regressor with interaction modeling (shared NN)

karl-richter:
**Status quo**
For each covariate (aka lagged regressor), a seperate net is trained.

**Change**
All lagged regressors share a network. In plot_parameters, the attribution of each covariate on the forecast is derived using the Captum model attribution method for deep networks.


alfonsogarciadecorral:
Hi @karl-richter it looks good to me in terms of model architecture changes

Screenshot 2022-11-06 at 16 22 31_903
Screenshot 2022-11-06 at 16 50 36_main

karl-richter:
Update: this pr is currently failing under the migration to Lightning, since Lightning abstracts the migration of tensors between accelerators and the CPU, the information required to calculate model attributions using Captum seems lost. 

The following exception is raised in line 91 of utils_torch.py when calculating the attributions using:
`target_attribution = saliency.attribute(model_input, target=[(output_feature, quantile)], abs=False)`

```python
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/karl/git/neural_prophet/venv/lib/python3.8/site-packages/captum/log/__init__.py", line 35, in wrapper
    return func(*args, **kwargs)
  File "/Users/karl/git/neural_prophet/venv/lib/python3.8/site-packages/captum/attr/_core/saliency.py", line 130, in attribute
    gradients = self.gradient_func(
  File "/Users/karl/git/neural_prophet/venv/lib/python3.8/site-packages/captum/_utils/gradient.py", line 119, in compute_gradients
    grads = torch.autograd.grad(torch.unbind(outputs), inputs)
  File "/Users/karl/git/neural_prophet/venv/lib/python3.8/site-packages/torch/autograd/__init__.py", line 275, in grad
    return Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
RuntimeError: element 0 of tensors does not require grad and does not have a grad_fn
```

My suspicion is that Lightning removes the information required to use autograd from the model output, which would be nessecary to calculate gradients.

alfonsogarciadecorral:
Hey @karl-richter 

What do you mean with

"Inspect plot_components of lagged regressors (seems to be always proportional for the sum of all regressors)"

karl-richter:
> Hey @karl-richter
> 
> What do you mean with
> 
> "Inspect plot_components of lagged regressors (seems to be always proportional for the sum of all regressors)"

Hi @alfonsogarciadecorral, in the plot_components, we have to distribute the joint covariate forecast to each of the covariates (one seperate plot of the forecast of each covariate). With the current method of distributing the attributions, it seems like each variable gets attributed evenly at each point. I will post an example later.

neuralprophet/forecaster.py
```diff
@@ -3089,7 +3096,8 @@ def _reshape_raw_predictions_to_forecst_df(self, df, predicted, components):
                         yhat = np.concatenate(([np.NaN] * pad_before, forecast, [np.NaN] * pad_after))
                         if j == 0:  # temporary condition to add only the median component
                             name = f"{comp}{forecast_lag}"
-                            df_forecast[name] = yhat
+                            yhat_df = pd.Series(yhat, name=name).set_axis(df_forecast.index)
+                            df_forecast = pd.concat([df_forecast, yhat_df], axis=1, ignore_index=False)
```
karl-richter, line 3100: [for reviewer]: this cause pandas warnings with the previous implementation. Now fixed.

neuralprophet/time_net.py
```diff
@@ -9,7 +10,17 @@
...
+from neuralprophet.utils import (
+    check_for_regularization,
+    config_events_to_model_dims,
+    config_regressors_to_model_dims,
+    config_seasonality_to_model_dims,
+    reg_func_events,
+    reg_func_regressors,
+    reg_func_season,
+    reg_func_trend,
+)
```
karl-richter, line 23: [for reviewer] importing utils directly caused an cyclic dependency error, thus we import all functions seperatly

alfonsogarciadecorral (changes requested):
Hi @karl-richter @ourownstory 

Nice one, I think this is a nice improvement!

A few things from me:

- `num_hidden_layers` - Maybe we can rename this to `num_hidden_layers_lagged_regressors` or, if want to share with ar_net, `num_hidden_layers_lagged`.  For example, Im building a shared network for future_regressors and I should be able to use a different configuration.
- Out of curiosity, how come you use that function/formula to create `d_hidden` if it is not passed by user? 
- Would it make sense to pass the neural network architecture as an array? Something like `covar_net_layers_array` which would replace current `n_hidden` and `num_hidden_layers`

alfonsogarciadecorral:
Hey @karl-richter  @ourownstory  

The commit I did was about removing `num_hidden_layer` and  `d_hidden` and instead initialiting the `ar_net` and `covar_net` networks through then new variables: `ar_net_layers_array` and  `covar_net_layers_array`

Have a look, because it implies a few changes:
- We remove all which was related with `num_hidden_layer` and  `d_hidden` . So now the user either specifies the network or it assumes `num_hidden_layers=0`. There is no option to pass the `num_hidden_layers` and then getting the `d_hidden` through a heuristic method.
- config_model --> this will no longer contain `num_hidden_layer` and  `d_hidden` and it will have the `covar_net_layers_array`
- I removed the possibilty of passing the ‘local’ network on method `add_lagged_regressor` as it is a shared network now.

If you agree on the new changes there are two things missing that I will do as soon as you give thumbs up
**IMPORTANT -MISSING:**
- Change tutorials documentation to remove the `num_hidden_layer` and  `d_hidden`
- Add an warn message saying that `num_hidden_layer` and  `d_hidden` is not used anymore

ourownstory (approved):
Excellent work @karl-richter and @alfonsogarciadecorral !!

ourownstory (approved):

PR #908, opened 2022-10-26
[logging] display logs in console

karl-richter:
**Status quo**
Currently all logs that are raised during the pytests are invisible.

**Change**
With this pr, logs that occur during the execution of a test are logged to cli (the debug console of the IDE).
For all logs from NeuralProphet, the log level is WARNING, for logs that are raised in the respective test, the log level is DEBUG.

Note: The amount of logs for some test might seem very high, but keep in mind that all logs that are raised during the tests are also displayed to the user in the end. Thus, we should be considerate of what to log on which level.


tests/__init__.py
```diff
@@ -0,0 +1,3 @@
+import logging
+
+logging.getLogger("matplotlib").setLevel(logging.WARNING)
```
karl-richter, line 3: matplotlib is known to spam logs, thus we set the logging level to warning.

karl-richter (commented):
added comment

karl-richter:
@ourownstory would be great if we could include this asap on the main branch as it allows to better debug logs.

tests/__init__.py
noxan: sounds good, might be great to have the same comment in the source code so people do not get rid of it, lacking context

PR #909, opened 2022-10-27
813 Remove regressor if training set has one unique value

### LINKED ISSUE #813: [enhancement]  remove regressor if training set has one unique value
Instead of throwing an error, simply remove the problematic regressor from the model with an error log message.
https://github.com/ourownstory/neural_prophet/blob/ea717ad31dec216ba670492f9b89fd075621dc3a/neuralprophet/df_utils.py#L318

This has caused issues for people in the past.
e.g. #809

This  must also address:
- reduce error to warning when unused column in df ( training or predicting have unused cols not in model definition)

karl-richter:
Removes regressor if only a unique value is present. Resolves #813


Kevin-Chen0 (approved):
LTGM

PR #914, opened 2022-10-27
[enhancement] add pytests for matplotlib 

LeonieFreisinger:
**Problem**
pytests in test_plotting() only cover plotly and do not cover matplotlib plotting backend.

**Solution**
Adapt all existing pytests to cover the plotly and matplotlib plotting backend. 

 

- [x]  My code follows the style guidelines described under CONTRIBUTING.md.
- [x]  I have added pytests to check whether my feature / fix works
- [x]  All gh actions pass or the failure is not related to my code.


LeonieFreisinger:
@karl-richter I transferred all plotting-related test functions to the test_plotting and parametrized both backends. Would be great if you can review the changes since you are well familiar with the plotting functionalities. Thank you!

@Kevin-Chen0 I marked this PR for the upcoming sprint.

karl-richter (approved):
Looks really good to me!
The only thing we should consider performance wise - does it really make sense to run the same test twice? since technically all parameters are the same and only the output is visualized using two frameworks we could also loop through the plottling backends within the tests. This should technically be faster. However, the current method provides more visibility. Not sure if you have discussed this with oskar yet.

Kevin-Chen0:
Hi Leonie, can you take a look and see why the model / metrics (pull_request) is failing? You can sync up with @judussoari and @karl-richter about this.

LeonieFreisinger:
> Hi Leonie, can you take a look and see why the model / metrics (pull_request) is failing? You can sync up with @judussoari and @karl-richter about this.

@Kevin-Chen0 the metric test currently fails in all my pull request. However, there is already a fix on the cml repo https://github.com/iterative/cml/pull/1228. As soon as this PR will be deployed, the error will resolve.

tests/test_plotting.py
```diff
@@ -24,10 +24,13 @@
 LR = 1.0
 
 PLOT = False
+# plot tests cover both plotting backends
```
Kevin-Chen0, line 27: I am wondering whether you can use `matplotlib.isinteractive()` from PR #960 instead of having a `PLOT` variable?

LeonieFreisinger: @Kevin-Chen0 Thanks for bringing up the idea. Yes, I can add the `matplotlib.isinteractive()`. Should we fully remove the PLOT variable? Then we would not have a manual switch to turn the plotting out/on anymore. Would it be an idea to add the `matplotlib.isinteractive()` to the if queries?

LeonieFreisinger: @karl-richter would be great to quickly have your opinion on it.

Kevin-Chen0: If we have `matplotlib.isinteractive()`, then we should remove the PLOT variables in the tests, as it is hardcoded as False by default for the test. @karl-richter thoughts?

LeonieFreisinger:
@karl-richter would be amazing if you could merge this PR. Please, wait for the PR#844 to be merged (by @noxan). In this order, we will have an easier time with merge conflicts. Thank you!

tests/test_plotting.py
```diff
@@ -398,45 +459,48 @@ def test_plotly_future_reg():
...
 
 
-def test_plotly_uncertainty():
-    log.info("testing: Plotting with plotly")
+@pytest.mark.parametrize(*decorator_input)
+def test_plot_uncertainty(plotting_backend):
+    log.info(f"testing: Plotting with uncertainty without forecast in focus with {plotting_backend}")
     df = pd.read_csv(PEYTON_FILE, nrows=NROWS)
 
     m = NeuralProphet(epochs=EPOCHS, batch_size=BATCH_SIZE, learning_rate=LR, quantiles=[0.9, 0.1])
```
Kevin-Chen0, line 477: @LeonieFreisinger Can you change the quantiles to `[0.25, 0.75]` for the `test_plotly_uncertainty()`? This is what I originally had in the test_uncertainty.py. You'll also need the change the `m.plot_parameters(quantile=0.75 ...)`, with the exception of the part in `with pytest.raises(ValueError)`. Thx

tests/test_plotting.py
```diff
@@ -398,45 +459,48 @@ def test_plotly_future_reg():
...
         fig1.show()
         fig2.show()
         fig3.show()
 
 
-def test_plotly_uncertainty():
-    log.info("testing: Plotting with plotly")
+@pytest.mark.parametrize(*decorator_input)
+def test_plot_uncertainty(plotting_backend):
+    log.info(f"testing: Plotting with uncertainty without forecast in focus with {plotting_backend}")
```
Kevin-Chen0, line 474: Change to:

`log.info(f"testing: Plotting with uncertainty estimation without forecast in focus with {plotting_backend}")`

Also, what do you mean by `without forecast in focus`?

tests/test_plotting.py
LeonieFreisinger: @Kevin-Chen0 thanks a lot for reviewing and for your feedback! 
I adjusted the quantiles accordingly.

tests/test_plotting.py
LeonieFreisinger: @Kevin-Chen0 I adjust the log.info messages as follows: 
`log.info(f"testing: Plotting with uncertainty estimation with {plotting_backend}")`
and
`log.info(f"testing: Plotting with uncertainty estimation with highlighted forecast step with {plotting_backend}")`
What do you think?

PR #934, opened 2022-11-04
Address small tech debt to use a conditional instead of try/except for Python 3.7 support

kapoor1992:


noxan:
Hi @kapoor1992 and many thanks for your contribution.

To me the two implementations are exactly the same and I got the impression the `try catch` with `ImportError` is a frequently seen pattern, e.g. [[stack overflow]](https://stackoverflow.com/questions/7604636/better-to-try-something-and-catch-the-exception-or-test-if-its-possible-first). Do you know of any other projects using the version check implementation? I'd prefer to follow the most common implementation unless there are any strong reasons for one over the other?

kapoor1992:
Hey @noxan! So, I found a lot more back-and-forth on this (even in your link) haha! The `try/except` pattern definitely follows the EAFP principle which is Pythonic in nature while the `if/else` pattern is deemed "safer" generally since you're checking first. Ultimately, I don't think either choice makes much of a difference here since we're not doing anything major other than picking an import. If you prefer `try/except`, then feel free to reject + close this PR :)

noxan (commented):
Looks good to me, thanks for the second opinion @karl-richter :)

Kevin-Chen0 (approved):
LGTM

Issue #939, opened 2022-11-09
Migration from prophet jupyter notebook does not build for docs


noxan:
The `Migration_from_Prophet` jupyter notebook fails while building docs (for now the nblink is removed so it works)

See failing builds [logs](https://github.com/ourownstory/neural_prophet/actions/runs/3374737674/jobs/5600672770).

After adding `prophet` as requirement to docs, the problem was that `neuralprophet` was not found - which is rather weird given it's certainly there. Installing `neuralprophet` within the github action (via `pip install -e .`) failed even worse.

PR #944, opened 2022-11-09
Migration from prophet jupyter notebook does fails docs build

noxan:
#939


ourownstory:
Will do another fix: remove prophet dependency in the existing tutorials.

ourownstory:
Will do another fix: remove prophet dependency in the existing tutorials.

leoniewgnr:
tackled in #1332

PR #946, opened 2022-11-12
Labels in CONTRIBUTING.md

karl-richter:


noxan:
I would strongly advocate for a bot to keep our labels on PRs updated. we just saw that in the last core meeting, github has already status indicators built in (e.g. needs review, approved, needs changes) and we should follow those for our labels. Otherwise we have mismatched information and rather cause confusion, my 2 cents. Happy to have a chat on that.

noxan (approved):
I would propose to merge the current state as we already use the labels and have anyone propose and discuss changes as follow up pull requests.

PR #955, opened 2022-11-14
[refactor] Run pyupgrade on project to update to latest python syntax

noxan:


noxan:
I'll close this for now as it seems to work with python 3.7 but not 100% convinced there wouldn't be any compatibility issues

Issue #928, opened 2022-11-02
[bug] Build tests / build (macos & windows-latest, 3.7 & 3.10) left hanging in PR #802


Kevin-Chen0:
In PR [#802](https://github.com/ourownstory/neural_prophet/pull/802), the following 4 tests ended up getting cancelled after taking forever to run:

![image](https://user-images.githubusercontent.com/13785749/199394143-c7f452c7-168b-4e4b-a756-f5be8284a88d.png)

This is caused because the following test is left hanging: `test_split_conformal_prediction`:

![image](https://user-images.githubusercontent.com/13785749/199394845-73e75031-516d-4a14-95d9-5d95736f5d85.png)

However, when run locally (at least on Windows), this test passes. Please advise.

Kevin-Chen0:
I have resolved the issue for Windows. It is because somewhere embedded in my code, there is a `plt.show()` being run, which caused the test case to hang.

PR #958, opened 2022-11-15
Added plot param of the NeuralProphet fit() and uncomment test_progress_display in test_integration.py.

Kevin-Chen0:
Added plot parameter in the NeuralProphet fit method that can toggle whether `plt.show()` is executed or not. Therefore, this param can be set to `False` for unit testing. This will prevent `test_progress_display` from hanging the PR tests, as seen Issue [#928](https://github.com/ourownstory/neural_prophet/issues/928).

### DIFF
--- a/neuralprophet/forecaster.py
+++ b/neuralprophet/forecaster.py
@@ -635,7 +635,7 @@ class NeuralProphet:
         self.config_season.append(name=name, period=period, resolution=fourier_order, arg="custom")
         return self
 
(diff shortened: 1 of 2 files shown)


karl-richter:
Hi @Kevin-Chen0 which system are you running on? Linux, correct? We should find someone who runs on Windows and let them test whether .show() works as expected on all systems.

karl-richter (approved):
I checked the behaviour manually again, looks good from my side

PR #960, opened 2022-11-15
refactored test failure

karl-richter:
I found a more elegant way on how we can avoide the pytest with .show to fail.
It seems like the cause of the issue is that the python session on the gh action workers is non-interactive, which causes the `.show()` call to crash the session. This can be avoided by calling `.isinteractive()` on the matplotlib session.

Let me know what you think.

References:
- https://stackoverflow.com/questions/29620884/how-to-put-python-3-4-matplotlib-in-non-interactive-mode
- https://gist.github.com/matthewfeickert/84245837f09673b2e7afea929c016904

### DIFF
--- a/neuralprophet/forecaster.py
+++ b/neuralprophet/forecaster.py
@@ -7,7 +7,7 @@ from typing import Optional, Union
 import numpy as np
 import pandas as pd
 import torch
-from matplotlib import pyplot as plt
+import matplotlib
 from torch.utils.data import DataLoader
 
 from neuralprophet import configure, df_utils, metrics, time_dataset, time_net, utils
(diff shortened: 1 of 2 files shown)


Kevin-Chen0:
This is failing in `test_wrapper_plots` for `windows-latest` OS. Can you check why?

Kevin-Chen0:
This is failing in `test_plot` in `test_integration.py` for `windows-latest, 3.10` OS. Can you check why? Is it b/c of a `plt.show()` being called?

`m.plot_latest_forecast(forecast, include_previous_forecasts=10)`

Kevin-Chen0 (approved):
LGTM

PR #420, opened 2021-10-14
GPU support

ourownstory:

PR #961, opened 2022-11-15
[lightning] GPU support

karl-richter:
Adds functionality to train the TimeNet model using an accelerator.
Supports GPU, MPS (M1) automatically and all other available Pytorch Lightning accelerators via manual declaration (eg. TPU etc.). Updated after Lightning merge.

Closes #420 
Closes #938


Kevin-Chen0:
Can you make test case(s) for this to improve the codecov?

karl-richter:
> Can you make test case(s) for this to improve the codecov?

@Kevin-Chen0 Note sure if that makes sense in this case, bc the machine would need to have a GPU / accelerator aside from CPU to test the feature. As far as I know thats not available / reasonable in gh actions.

Kevin-Chen0:
Can you resolve the merge conflict first?

Kevin-Chen0:
Karl, can you address the two questions I have before I approve? Thx.

noxan (approved):
Code looks good to me and no merge conflicts.

I agree with @Kevin-Chen0 that it would be great to have test cases, yet as @karl-richter pointed out there is no option with github actions I'd be aware of - if someone knows any good way lets create test cases with a follow up issue 👍

geneman:
running with CUDA does not work: RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cpu and cuda:0!

PR #966, opened 2022-11-16
Renamed plot_forecast.py and plot_model_parameters.py to xxx_matplotlib.py

AdityaNahata:
Renamed plot_forecast.py and plot_model_parameters.py to plot_forecast_matplotlib.py and plot_model_parameters_matplotlib.py


karl-richter:
@AdityaNahata Looks like you need to update the imports in the forecaster.py and potentially also other files. fyi - the metrics job does currently not yet work on forks, thus we do not have to make it pass for our pr. However, all other tests should pass.

AdityaNahata:
@karl-richter Thanks for the feedback. This is my first PR, so sorry for the problems raised. I will be working on the fix soon.

PR #974, opened 2022-11-17
[feature] Add a main script to print neuralprophet version

noxan:
Allows users to run `python -m neuralprophet --version` and receive the version information of the package directly from their terminal. Could be extended in the future to a cli application of neuralprophet.


ourownstory (changes requested):
Nice addition!
Let's just use the long flag "--version".
Let's remove "-v", which is often used for verbose

noxan:
@ourownstory Good point. Should we go with `-V` for version (capital letter) or fully remove the short version?

neuralprophet/__main__.py
```diff
@@ -0,0 +1,16 @@
...
+Invokes neuralprophet when module is run as a script.
+"""
+import argparse
+
+from neuralprophet._version import __version__
+
+
+def parse_args(args=None):
+    parser = argparse.ArgumentParser(description="NeuralProphet")
+    parser.add_argument("-v", "--version", action="version", version="%(prog)s " + __version__)
```
noxan, line 11: Remove `-v` option

```suggestion
    parser.add_argument("--version", action="version", version="%(prog)s " + __version__)
```

noxan, line 11: Use capital letter `-V` for version shorthand

```suggestion
    parser.add_argument("-V", "--version", action="version", version="%(prog)s " + __version__)
```

karl-richter (approved):

PR #983, opened 2022-11-23
[tests] Add tests for configure

noxan:
Basic tests for training config quantiles initialization.

Motivation: if we refactor the handling of configuration options and their internals it is good to have some test cases to ensure we do not break anything. This is a starting point for more work in this direction and to provide the template for other test cases people can replicate if they work on some of the configuration objects and want some safety net :)


karl-richter (approved):
Looks good, 🎂  for more tests!

PR #987, opened 2022-11-24
[typing] Resolve type annotation issues in configure file

noxan:
Resolve type annotation issues in configure file and ignore epochs `None` type warning due to almost impossible to fix with `dataclasses` - see [[source]](https://stackoverflow.com/questions/56839030/dataclasses-how-to-ignore-none-values-using-asdict)

### DIFF
--- a/neuralprophet/configure.py
+++ b/neuralprophet/configure.py
@@ -266,7 +266,7 @@ class Trend:
 
         if self.trend_reg_threshold is None:
             pass
-        elif type(self.trend_reg_threshold) == bool:
+        elif isinstance(self.trend_reg_threshold, bool):
             if self.trend_reg_threshold:
                 self.trend_reg_threshold = 3.0 / (3.0 + (1.0 + self.trend_reg) * np.sqrt(self.n_changepoints))
                 log.debug(f"Trend reg threshold automatically set to: {self.trend_reg_threshold}")
--- a/neuralprophet/configure.py
+++ b/neuralprophet/configure.py
@@ -6,7 +6,7 @@ import math
 import types
 from collections import OrderedDict
 from dataclasses import dataclass, field
-from typing import Callable, List, Optional
+from typing import Callable, List, Optional, Type
 from typing import OrderedDict as OrderedDictType
 from typing import Union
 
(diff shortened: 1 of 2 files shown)


neuralprophet/configure.py
```diff
@@ -226,7 +226,7 @@ def set_lr_finder_args(self, dataset_size, num_batches):
         )
 
     def get_reg_delay_weight(self, e, iter_progress, reg_start_pct: float = 0.66, reg_full_pct: float = 1.0):
-        progress = (e + iter_progress) / float(self.epochs)
+        progress = (e + iter_progress) / float(self.epochs)  # type: ignore
```
karl-richter, line 229: is this `# type: ignore` an artifact? @noxan

neuralprophet/configure.py
```diff
@@ -263,7 +263,9 @@ def __post_init__(self):
             self.n_changepoints = len(self.changepoints)
             self.changepoints = pd.to_datetime(self.changepoints).sort_values().values
 
-        if type(self.trend_reg_threshold) == bool:
+        if self.trend_reg_threshold is None:
+            pass
+        elif type(self.trend_reg_threshold) == bool:
```
karl-richter, line 268: shouldn't this be `isinstance(..., bool)`?

neuralprophet/configure.py
```diff
@@ -285,7 +287,7 @@ def __post_init__(self):
             else:
                 log.info("Trend reg lambda ignored due to no changepoints.")
                 self.trend_reg = 0
-                if self.trend_reg_threshold > 0:
+                if self.trend_reg_threshold and self.trend_reg_threshold > 0:
```
karl-richter, line 290: what exactly is this code supposed to do? why is > 0 not sufficient?

karl-richter (changes requested):
Can you check the comments in configure.py? Otherwise lgtm

neuralprophet/configure.py
noxan: it's intentional as there is no real way to avoid the potential `None` type for epochs as we otherwise couldn't use dataclasses (they do not differ between type annotations for the attribute and the constructor).

neuralprophet/configure.py
noxan: good point, did not change it to my defense, haha ;)

neuralprophet/configure.py
noxan: yes, it could also be `None` which would fail if we compare `None > 0`

noxan: Screenshot 2022-11-29 at 6 00 16 PM

neuralprophet/configure.py
noxan: fixed

neuralprophet/configure.py
noxan: added a comment on it, would be great to resolve it at some point in time - would sadly for now result in a lot of boilerplate code - read me [here](https://stackoverflow.com/questions/64670320/how-to-avoid-checking-for-none-when-setting-optional-dataclass-args-in-post-in)

karl-richter (approved):
lgtm

PR #995, opened 2022-11-28
Time in model training benchmark

karl-richter:
Added training time to the benchmarks + only hide the figures, but not the metrics.

### DIFF
--- a/.github/workflows/metrics.yml
+++ b/.github/workflows/metrics.yml
@@ -39,12 +39,12 @@ jobs:
       - name: Compare performance
         run: |
           echo "## Model Benchmark" >> report.md
-          echo "<details>\n<summary>Show benchmark results</summary>\n" >> report.md
           python tests/metrics/compareMetrics.py >> report.md
       - name: Publish report
         env:
           REPO_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
+          echo "<details>\n<summary>Model training plots</summary>\n" >> report.md
           echo "## Model Training" >> report.md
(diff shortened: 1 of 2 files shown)


karl-richter:
@Kevin-Chen0 @ourownstory @noxan 
Can someone review and merge this pr asap pls? Need the time in the benchmarks to check that the Lightning speedup works.

.github/workflows/metrics.yml
```diff
@@ -39,12 +39,12 @@ jobs:
       - name: Compare performance
         run: |
           echo "## Model Benchmark" >> report.md
-          echo "<details>\n<summary>Show benchmark results</summary>\n" >> report.md
           python tests/metrics/compareMetrics.py >> report.md
       - name: Publish report
         env:
           REPO_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
+          echo "<details>\n<summary>Model training plots</summary>\n" >> report.md
```
noxan, line 47: I'd prefer to have an option to also collapse the regular metrics - no hard feelings and no blocker for sure 👍

noxan (approved):
Looks good to me, one minor comment on collapsible metrics - ready to merge from my side.

.github/workflows/metrics.yml
karl-richter: Totally, agree, in the current setting there are too many metrics. Lets agree in our weekly meeting on the few key metrics that we always display and make all the others collabsable. I am just afraid that people do not check their metrics if its not visible that their pr reduces performance.

PR #997, opened 2022-11-28
Meta name bool

alfonsogarciadecorral:
@ourownstory 

This simple PR is to reduce some lines the global-local logic.

A boolean variable called `meta_name_bool` will be used if we need to know the time series ID when we interact with the Model.

Which is the best release I should add this to?


karl-richter (approved):
Love the change, thanks for simplifying the implementation!

neuralprophet/forecaster.py
```diff
@@ -705,6 +705,9 @@ def fit(self, df, freq="auto", validation_df=None, progress="bar", minimal=False
         # When only one time series is input, self.id_list = ['__df__']
         self.num_trends_modelled = len(self.id_list) if self.config_trend.trend_global_local == "local" else 1
         self.num_seasonalities_modelled = len(self.id_list) if self.config_season.global_local == "local" else 1
+        self.meta_name_bool = (
+            False if (self.num_trends_modelled == 1 and self.num_seasonalities_modelled == 1) else True
```
noxan, line 709:
```suggestion
            not (self.num_trends_modelled == 1 and self.num_seasonalities_modelled == 1)
```

noxan, line 709:
```suggestion
            self.num_trends_modelled != 1 or self.num_seasonalities_modelled != 1
```
Should do not the same just easier to read

neuralprophet/forecaster.py
```diff
@@ -705,6 +705,9 @@ def fit(self, df, freq="auto", validation_df=None, progress="bar", minimal=False
         # When only one time series is input, self.id_list = ['__df__']
         self.num_trends_modelled = len(self.id_list) if self.config_trend.trend_global_local == "local" else 1
         self.num_seasonalities_modelled = len(self.id_list) if self.config_season.global_local == "local" else 1
+        self.meta_name_bool = (
```
noxan, line 708: Not really sure what `meta_name_bool` stands for, could we come up with some more clear naming? For example `meta_use_df_name` or something similar?

noxan (changes requested):
LGTM overall, like the removal of duplicate code. I've added two suggestions, should be quick fixes for better readability - we can merge it with any upcoming release.

neuralprophet/forecaster.py
karl-richter: I agree, for me its also not really clear what meta_name is referring to.

neuralprophet/forecaster.py
alfonsogarciadecorral: Thank you @noxan !

neuralprophet/forecaster.py
alfonsogarciadecorral: Thanks, I agree!
What about:
- meta_used_in_forward
- model_uses_meta
?

karl-richter: What exactly is meta even referring to? Its the name of the local dataset in case global / local modelling is used, correct?

alfonsogarciadecorral: Yes, the dataloader has meta in it, including the 'ID' of the local dataset.
When doing global/local modeling we need to know the 'ID' of each sample on the batch.

alfonsogarciadecorral: changed it to  `meta_used_in_model` - Didn't add "name" on it as this could be used as well once we have static covariates, for example

noxan: Thanks for the improvement @alfonsogarciadecorral - I find the `meta` naming of this functionality rather unclear, but that is not related to your work - I'll open a follow up issue on that.

noxan (approved):
LGTM - create the follow up issue separately

Issue #763, opened 2022-09-07
Add support for multiple countries holidays


noxan:

leoniewgnr:
Can I work on this?

noxan:
@leoniewgnr sure, thanks for the pull request already :)

Issue #712, opened 2022-08-11
Harmonize internal event and holiday implementation


noxan:
Holidays are only a special kind of events, yet events and holidays often have two almost the same implementations. This issue's aim is to harmonize the implementations and only have a single generic implementation, removing almost duplicate code.

More information and details to follow.

leoniewgnr:
Can I work on this?

leoniewgnr:
This is done:) Would be great if anyone could link the pull request #1110 to this issue

PR #1001, opened 2022-11-29
Support for holidays of multiple countries

leoniewgnr:
👨‍🔬 Status quo
Currently it is only possible to consider holidays of one country. If another country is added, the previous one is overwritten. Addresses issue #763 

👨‍🎨 Key changes
It is possible now to add a list of countries, whose holidays should be considered. For example m.add_country_holidays(country_name = ["US", "Germany"]) will consider holidays of the US and Germany. For redundant holidays (here: 'Christmas Day' and 'Erster Weihnachtstag') the name of the last mentioned country will be kept (here: 'Erster Weihnachtstag')

🕵️‍♂️ Review Checklist
- [x] I have commented my code, added Docstrings and data types to function definitions.
- [x] My code follows the style guidelines described under CONTRIBUTING.md.
- [x] I have added pytests to check whether my feature / fix works
- [ ] All gh actions pass or the failure is not related to my code.


neuralprophet/utils.py
```diff
@@ -279,8 +279,8 @@ def get_holidays_from_country(country, df=None):
 
     Parameters
     ----------
-        country : str
-            Country name to retrieve country specific holidays
+        country : list
+            List of country name sto retrieve country specific holidays
```
noxan, line 283:
```suggestion
            List of country names to retrieve country specific holidays
```

neuralprophet/utils.py
```diff
@@ -294,16 +294,22 @@ def get_holidays_from_country(country, df=None):
...
+    if isinstance(country, str):
+        country = [country]
 
-    try:
-        with warnings.catch_warnings():
-            warnings.simplefilter("ignore")
-            holiday_names = getattr(hdays_part2, country)(years=years).values()
-    except AttributeError:
+    holidays = {}
+    for c in country:
```
noxan, line 302: Would be great for readability to use a more verbose name than `c` for country, we should consider renaming the `country` parameter to `country_list` or something to make `country` free for the individual country or go with something like `single_country`

tests/test_unit.py
```diff
@@ -992,3 +992,19 @@ def test_add_country_holiday_multiple_calls_warning(caplog):
...
+        batch_size=BATCH_SIZE,
+        learning_rate=LR,
+    )
+    m.add_country_holidays(country_name = ["US", "Germany"])
+    metrics = m.fit(df, freq="D")
+    forecast = m.predict(df)
+    #get the name of holidays and compare that no holiday is repeated
+    holiday_names = m.model.config_holidays.holiday_names
+    assert 'Christmas Day' not in holiday_names
+    assert 'Erster Weihnachtstag' in holiday_names
```
noxan, line 1010: Love that you added a test right away, big props. Might make sense to test for a few more holidays from both countries, currently the test case would not fail if only holidays from `Germany` were added.

noxan (changes requested):
Great work, looks very good to me overall. I've added a few minor suggestions and there are a few flake8 warnings which you should look at (one of them is actually critical as the variable name is not defined). Nothing big, you're on a good track, keep it up 👍

neuralprophet/utils.py
leoniewgnr: Thanks, I went for `single_country` as I wanted to keep country to keep it consistence with other functions' naming

tests/test_unit.py
leoniewgnr: Good point! Thanks!

neuralprophet/forecaster.py
```diff
@@ -590,8 +590,8 @@ def add_country_holidays(self, country_name, lower_window=0, upper_window=0, reg
 
         Parameters
         ----------
-            country_name : string
-                name of the country
+            country_name : string, list
```
noxan, line 593:
```suggestion
            country_name : str, list
```

neuralprophet/utils.py
```diff
@@ -279,8 +279,8 @@ def get_holidays_from_country(country, df=None):
 
     Parameters
     ----------
-        country : str
-            Country name to retrieve country specific holidays
+        country : list
```
noxan, line 281:
```suggestion
        country : str, list
```

neuralprophet/time_dataset.py
```diff
@@ -455,22 +455,29 @@ def make_country_specific_holidays_df(year_list, country):
...
+    for single_country in country:
         try:
-            country_specific_holidays = getattr(hdays_part1, country)(years=year_list)
+            single_country_specific_holidays = getattr(hdays_part2, single_country)(years=year_list)
         except AttributeError:
-            raise AttributeError(f"Holidays in {country} are not currently supported!")
+            try:
+                single_country_specific_holidays = getattr(hdays_part1, single_country)(years=year_list)
+            except AttributeError:
+                raise AttributeError(f"Holidays in {single_country} are not currently supported!")
```
noxan, line 478: Might be good to extract this code (as we have it multiple times) into a helper function? Also happy to have this as a follow up issue.

noxan (approved):
LGTM - I've added two minor suggestions which I'd add before merging and one note about duplicate code which we could extract to a helper function

neuralprophet/time_dataset.py
leoniewgnr: Good point! I think it would make sense to do a follow up issue and consider that together with issue #712 as there might be synergies effects

noxan: 👍

noxan (approved):

PR #1003, opened 2022-11-30
Update progress bar only every epoch

karl-richter:
Currently the progress bar by Lightning is supposed to be updated every step. With this pr, the progress bar is only updated once every epoch and shows the progress across all epochs.


noxan (approved):
Code looks good to me - functionality I tested in terms of it has a working progress bar 😅

karl-richter:
@noxan Tested the progress bar also with a validation dataset and everything seems to work fine.

PR #1005, opened 2022-11-30
Custom early stopping for OneCycleLR

karl-richter:
The standard early stopping configuration in Lightning is not compatible with the OneCycleLR, which increases the learning rate until a defined threshold (eg. 30%) and only afterwards decreases to find a global minima. The early stopping callback should be deactivated until the threshold is reached.

### DIFF
--- a/neuralprophet/utils.py
+++ b/neuralprophet/utils.py
@@ -845,7 +845,23 @@ def configure_trainer(
 
     # Early stopping monitor
     if config_train.early_stopping:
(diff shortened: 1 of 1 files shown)


karl-richter:
We need some experiments to find a good heuristic when to trigger the early stopping.

karl-richter:
We need to check again whether this change is needed, but potentially its a feature we do not desire.

ourownstory:
Closing as it introduces more potential issues than potential improvements due to complexity to the user.

Issue #998, opened 2022-11-28
Add deprecation warning for matplotlib


noxan:
Make sure people see a deprecation warning about the migration to plotly whenever they use matplotlib (explicit or implicit as default).

noxan:
Adjust https://github.com/ourownstory/neural_prophet/pull/984/files to show the warning always, not only after an explicit call.

PR #1006, opened 2022-11-30
Change plotting_backend deprecation warning for implicit and explicit matplotlib use

LeonieFreisinger:
referring Issue: #998

Key Changes: 
- outsourced warning message in plot_utils.py
- implement warning message to be shown for matplotlib on every plot function that the user can call. 

Review Checklist
- [x]  I have performed a self-review of my own code.
- [x]  My code follows the style guidelines described under CONTRIBUTING.md.
- [x]  I have added pytests to check whether my feature / fix works
- [x]  All gh actions pass or the failure is not related to my code.


LeonieFreisinger:
@noxan changed the warning to be displayed for any matplotlib plotting. Would be great if you can review it. P.S. I allocated the warning message in the plot_utils.py, since it is used multiple times. For having an overview of warning messages, I think it would be generally nice to have them all in one place, e.g. in a utils.py. However, it might as well be a little overkill if some cases have a warning message that is only used once. What do you think?

noxan (approved):
Awesome, so little code - love how efficiently you tackled the issue 👏 

We might need to resolve the merge conflict, if you find time would be awesome, otherwise I'll check tomorrow.

neuralprophet/plot_utils.py
```diff
@@ -10,6 +10,15 @@
 log = logging.getLogger("NP.plotting")
 
 
+
+def log_warning_deprecation_plotly(plotting_backend):
```
noxan, line 14:
```suggestion
def log_warning_deprecation_plotly(plotting_backend):
```

PR #1010, opened 2022-11-30
Move tensors to correct device

### LINKED ISSUE #1002: GPU support tensors
NeuralProphet does seem to raise an error when training on GPU's.

        running with CUDA does not work: RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cpu and cuda:0!

_Originally posted by @geneman in https://github.com/ourownstory/neural_prophet/issues/961#issuecomment-1331144742_

karl-richter:
Resolves #1002

Initially we assumed that with Lightning we do not have to worry about moving tensors to devices at all, but it turns out that this assumption only holds for tensor created via the data loaders. All manually created tensors need to be moved to the correct device (which is very straight forward in Lightning due to the self.device arg, [Reference](https://pytorch-lightning.readthedocs.io/en/stable/starter/converting.html#remove-any-cuda-or-to-device-calls)).

Adds a pytest that uses an accelerator if available. On github action workers, accelerators are most likely not available, but on local machines this could help detect issues early.


noxan (approved):
LGTM - not sure if there are any other tensors for which we need to specify the device, yet anyhow better to have more of them on the right device 👍

PR #1020, opened 2022-12-02
[refactor] Remove duplicate season reg lambda

noxan:
Remove the `config_train.reg_lambda_season` as it is only a duplicate of `config_season.reg_lambda`


judussoari (approved):
lgtm

PR #1022, opened 2022-12-02
[fix] Replace all deprecated plot_last_forecast with plot_latest_forecast

noxan:
Thanks @gozdii for the hint in https://github.com/ourownstory/neural_prophet/discussions/890


judussoari (approved):
LGTM, it's important to be consistent with our func namings

PR #1028, opened 2022-12-03
[devops] Remove CodeQL Github CI job

noxan:
## :microscope: Background

- There are a lot of Github CI jobs running for each commit and it starts to get slow
- Nobody seems to be really using CodeQL - or does anyone?

## :crystal_ball: Key changes

- Remove CodeQL Github CI job as it is not used


karl-richter:
Idea: We could only run this job on the main branch, this way we still have the check but it does not run for every push to every single pr.

noxan:
@karl-richter love the idea :)

noxan:
Follow up in #1031

PR #1038, opened 2022-12-05
[feature] Customized forecast frequency

judussoari:
If the user opts to specify `prediction_frequency`, the forecasting procedure of NP will be adjusted. 

**Before:**
Dealing with a dataset of daily data resolution and specified `n_forecasts = 7,` NeuralProphet forecasts 1 week ahead `yhat1..yhat7` on every day, matching the data resolution. 

**New**
With `prediction_frequency = 14` and `n_forecasts = 7`, a forecast for the next 7 days will be made every 14 days (instead of every day). 
In this case, only every 14th sample will be fitted and predicted, which might save us computation time (and maybe even increase accuracy?) 
If the user opts to choose different values for `n_forecasts` and `prediction_frequency`, trend and seasonality cannot be flattened out completely. 

First attached Screenshot: `forecast` df (output of `m.predict()`) as usual, without specifying `prediction_frequency`. 
Second screenshot:  `forecast` df with a specified `prediction_frequency = 7`

**Open ToDos**
- [x] Figure out what to do with trend and seasonality in this case
- [x] name of the arg forecast_period
- [x] dedicated tests with different combinations
- [x] Simplify and encapsulate by moving the striding logic to it's own method of TimeDataset

Bildschirmfoto 2022-12-04 um 21 55 38

Bildschirmfoto 2022-12-04 um 21 54 22


neuralprophet/forecaster.py
```diff
@@ -768,6 +768,13 @@ def predict(self, df, decompose=True, raw=False):
                 Options
                     * (default) ``False``: returns forecasts sorted by target (highlighting forecast age)
                     * ``True``: return the raw forecasts sorted by forecast start date
+            fcst_time : int
+                specifies the time at which a forecast shall start, instead of right at the end of the input data
+
+                Note
+                ----
+                If e.g. fcst_time=4 and n_forecasts=7, forecasts start at the 4th date after the last date of df, and
+                forecasts are only made every 7th step (once in a week in case of daily resolution).
```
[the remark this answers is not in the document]
ourownstory: we could parameterize this with start_date and repeat_interval (or similar)

neuralprophet/time_dataset.py
```diff
@@ -403,7 +408,7 @@ def _stride_lagged_features(df_col_name, feature_dims):
             tabularized_input_shapes_str += f"    {key} {value.shape} \n"
     log.debug(f"Tabularized inputs shapes: \n{tabularized_input_shapes_str}")
 
-    return inputs, targets, config_missing.drop_missing
+    return df, inputs, targets, config_missing.drop_missing # inputs, targets, und auch df filtern und zurueckgeben. gibt wahrsicbelich errors though.
```
noxan: 🇩🇪 ?

noxan (approved):
LGTM, two minor adjustments in comments would be great @judussoari

neuralprophet/time_dataset.py
```diff
@@ -377,13 +384,21 @@ def _stride_lagged_features(df_col_name, feature_dims):
...
 
         inputs["events"] = events
-
     if predict_mode:
         targets = np.empty_like(time)
         targets = np.nan_to_num(targets)
     else:
         targets = _stride_time_features_for_forecasts(df["y_scaled"].values)
 
+    if forecast_period is not None:
```
ourownstory, line 393: I wonder if we could make this a dedicated method to `TimeDataset`?
(called after `init_after_tabularized`)
Or, we could add it to `init_after_tabularized` to avoid another loop. 

Both methods would keep the already massive tabularization function leaner.

What do you think @judussoari @noxan ?

tests/test_integration.py
```diff
@@ -305,6 +305,19 @@ def test_ar():
...
     forecast = m.predict(df=future)
+    log.info("testing: AR only forecasting once per week instead of once per day")
+    m = NeuralProphet(
+        n_forecasts=7,
+        n_lags=7,
+        yearly_seasonality=False,
+        epochs=EPOCHS,
+        batch_size=BATCH_SIZE,
+        learning_rate=LR,
+        forecast_period=7,
```
ourownstory, line 316: I think this functionality deserves its dedicated test.
I think we should also test for mismatched (overlapping and non-overlapping) `n_forecasts` - `forecast_period`  combinations, 
e.g. 
overlapping: 7-1, 7-3
matched 7-7
with gaps: 1-7, 3-7

tests/test_unit.py
```diff
@@ -679,6 +679,8 @@ def test_globaltimedataset():
         m.config_normalization = config_normalization
         df_global = m._normalize(df_global)
         dataset = m._create_dataset(df_global, predict_mode=False)
+        # fcst_time is normally created within m.predict() and passed to the dataset class
```
ourownstory, line 682: is this the `start_time`?
I wonder if this needs to be exposed or documented?

neuralprophet/time_dataset.py
```diff
@@ -211,6 +211,7 @@ def tabularize_univariate_datetime(
     config_lagged_regressors: Optional[configure.ConfigLaggedRegressors] = None,
     config_regressors: Optional[configure.ConfigFutureRegressors] = None,
     config_missing=None,
+    forecast_period=None,
```
ourownstory, line 214: Please see my comment in `forecaster.py`

neuralprophet/forecaster.py
```diff
@@ -295,6 +295,12 @@ class NeuralProphet:
             Provide `None` to deactivate the use of accelerators.
         trainer_config: dict
             Dictionary of additional trainer configuration parameters.
+        forecast_period : int
```
ourownstory, line 298: `forecast_period` is often used as synonym for what we have called `n_forecasts`, so this could be ambiguous to many. 
I am having a hard time coming up with a better name. Here are some ideas:
- forecast_stride
- forecast_skip_period
- forecasting_frequency
- forecast_repeat_frequency

tests/test_unit.py
ourownstory: how is the time/start of the forecasting repeat frequency set?

ourownstory (changes requested):
Generally this looks good and is near ready for merge!

There are few things that I think we should address now:
* name of the arg `forecast_period`
* dedicated tests with different combinations
* how is the start-time set? (as n_lags, and CV splits affect where the first sample starts, this is hard to be controlled implicityly)

Optional/next PR: 
*Simplify and encapsulate by moving the striding logic to it's own method of TimeDataset

neuralprophet/time_dataset.py
```diff
@@ -158,6 +159,32 @@ def init_after_tabularized(self, inputs, targets=None):
...
+            forecast_frequency : int
+                periodic interval in which forecasts should be made.
+
+            Note
+            ----
+            E.g. if forecast_frequency=7, forecasts are only made on every 7th step (once in a week in case of daily resolution).
+        """
+        if forecast_frequency is None:
+            return
+        for key, data in self.inputs.items():
```
ourownstory, line 179: I wonder if we could do something like
`self.samples = self.samples[::forecast_frequency]`
Does that not work here?

neuralprophet/time_dataset.py
```diff
@@ -158,6 +159,32 @@ def init_after_tabularized(self, inputs, targets=None):
...
+        Parameters
+        ----------
+            forecast_frequency : int
+                periodic interval in which forecasts should be made.
+
+            Note
+            ----
+            E.g. if forecast_frequency=7, forecasts are only made on every 7th step (once in a week in case of daily resolution).
+        """
+        if forecast_frequency is None:
```
ourownstory, line 177: You could also add the case of `forecast_frequency==1` here

neuralprophet/forecaster.py
ourownstory: @noxan Could you please give some quick feedback: which one would you pick?

ourownstory (approved):
Great work! I think we can merge now.
I added two suggestions for potential quick improvements.

Further, I think that we should address (maybe in a subsequent PR) the starting time setting and dealing with missing data.

- how is the start-time set? (as n_lags, and CV splits affect where the first sample starts, this is hard to be controlled implicityly)
- Further, could there be a case where missing dates are not added, and thus gaps exists? Then, the approach to filter but every `forecast_frequency`th value would fail as it would get shifted by the gaps.
- An approach that does some sort of time modulo operation (if feasible), would likely address both issues.

neuralprophet/time_dataset.py
judussoari: Yes it does. Changed it, thanks for the suggestion!

neuralprophet/time_dataset.py
judussoari: good catch, thanks!

neuralprophet/time_dataset.py
judussoari: Great suggestion, did that!

ourownstory (approved):
Perfect!
Feel free to merge after rename to `prediction_frequency`

PR #1039, opened 2022-12-05
Refactored fit() interface

karl-richter:
## :microscope: Background

- We want to provide a user with the ability to activate / deactivate some of the non-required features during training. This includes checkpointing, logging of progress via the progress bar as well as the calculation of metrics during training.
- In the initial version of NeuralProphet, the `minimal` used a separate function to `_train()` called `_train_minimal()` that did not contain all of the above mentioned features and was "minimal".
- With the transition to Lightning, the training procedure is abstracted into the model and a lot simpler. Thus, we removed the `_train()` function and added `minimal` as a parameter to it.
- With the performance optimization of Lightning, the demand raised to control the training features individually. Thus, this PR supports controlling this behavior more granularly by using the `progress`, `metrics` and `checkpointing` parameter. The `minimal` mode further persists, but only remains a shortcut to set the other parameters to false.

## :crystal_ball: Key changes

- Added a deprecation warning for the `collect_metrics` parameter in `__init__` since this information can only be provided in the fit method in the future
- Added the `metrics` and `checkpointing` parameter to the `fit()` and subsequent functions
- Small refactorings and clean ups


karl-richter:
Hey @ourownstory @noxan, this PR would bring some interface changes (remove minimal, add metrics and checkpointing). Should we provide the metrics and checkpointing arg in NeuralProphet or in the fit() function?

noxan:
@karl-richter I like your approach, also agree with moving the metrics and checkpoints to where they actually matter (in the fit function) - would also make it very clear if people disabled metrics that they cannot expect that as return value.

karl-richter:
**Interface Change**
```python
# 0.4.2
def fit(
        self,
        df,
        freq="auto",
        validation_df=None,
        progress="bar",
        minimal=False
)

# 0.5.0
def fit(
        self,
        df: pd.DataFrame,
        freq: str = "auto",
        validation_df: pd.DataFrame = None,
        epochs: int = None,
        batch_size: int = None,
        learning_rate: float = None,
        early_stopping: bool = False,
        minimal: bool = False,
        metrics: bool = None,
        progress="bar",
        checkpointing: bool = False,
        continue_training: bool = False,
    )
```

karl-richter:
Added preparation for distributed training on request by @judussoari. This includes the `num_workers` param in `fit()`

noxan:
@karl-richter Many thanks for the great work. I'm still working on the review. For the next pull requests it would be great to split this into several smaller ones, as it is hard to review everything at once and especially to keep track which changes relate to what...

1. Replace minimal mode by checkpoints and metrics
2. Refactor fit method to allow continue training with other parameters
3. Prepare support for `num_workers`
4. A changed checkpoints implementation
5. Moving the progress bar implementation
6. ...

Happy to have a chat with you to explain this in detail, I'll try to conclude the review soon :)

neuralprophet/forecaster.py
```diff
@@ -654,7 +656,22 @@ def add_seasonality(self, name, period, fourier_order):
...
+        self,
+        df: pd.DataFrame,
+        freq: str = "auto",
+        validation_df: pd.DataFrame = None,
+        epochs: int = None,
+        batch_size: int = None,
+        learning_rate: float = None,
+        early_stopping: bool = False,
+        minimal: bool = False,
+        metrics: bool = None,
```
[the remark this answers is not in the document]
noxan: Or is it a special way of using the previously defined metrics? I mean that is a pretty confusing way if people want to explicitly define this 🤔

neuralprophet/forecaster.py
```diff
@@ -726,18 +792,30 @@ def fit(self, df, freq="auto", validation_df=None, progress="bar", minimal=False
...
+            )
         else:
             df_val, _, _, _ = df_utils.prep_or_copy_df(validation_df)
             df_val = self._check_dataframe(df_val, check_y=False, exogenous=False)
             df_val = self._handle_missing_data(df_val, freq=self.data_freq)
-            metrics_df = self._train(df, df_val=df_val, minimal=minimal, continue_training=continue_training)
+            metrics_df = self._train(
+                df,
+                df_val=df_val,
+                progress=bool(progress),
```
noxan, line 813: Same as with metrics comment above, we should have clear variable naming if they contain different values in different places

neuralprophet/utils.py
```diff
@@ -833,6 +837,6 @@ def _update_n(self, bar, value: int) -> None:
     config["enable_model_summary"] = False
     # TODO: Disabling sampler_ddp brings a good speedup in performance, however, check whether this is a good idea
     # https://pytorch-lightning.readthedocs.io/en/stable/common/trainer.html#replace-sampler-ddp
-    config["replace_sampler_ddp"] = False
+    # config["replace_sampler_ddp"] = False
```
noxan, line 840: Is this an intentional change?

noxan:
@karl-richter Overall a great set of changes. Might be best to have a short call on this. We should consider extracting changes into separate PRs and then review and merge them individually.

neuralprophet/forecaster.py
```diff
@@ -319,7 +317,6 @@ def __init__(
         ar_reg: Optional[float] = None,
         learning_rate: Optional[float] = None,
         epochs: Optional[int] = None,
-        early_stopping: bool = False,
```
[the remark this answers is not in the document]
karl-richter: Was never released in a public version, thus this should be fine

neuralprophet/forecaster.py
karl-richter, line 669:
```suggestion
        metrics: np_types.CollectMetricsMode = None,
```

neuralprophet/forecaster.py
```diff
@@ -701,6 +758,15 @@ def fit(self, df, freq="auto", validation_df=None, progress="bar", minimal=False
...
                 )
 
+        if progress == "plot" and metrics is False:
+            log.warning("Progress plot requires metrics to be enabled. Enabling the default metrics.")
+            metrics = metrics.get_metrics(True)
+
+        if minimal:
+            checkpointing = False
+            self.metrics = False
+            progress = None
```
karl-richter, line 768: Consider removing the minimal parameter or add a warning in case any other configuration (eg. checkpointing etc) is provided together with minimal.

neuralprophet/time_net.py
```diff
@@ -203,7 +199,8 @@ def __init__(
         self.batch_size = self.config_train.batch_size
 
         # Metrics Config
-        if not minimal:
+        self.collect_metrics = bool(metrics)  # yields True if metrics is not an empty dictionary
+        if self.collect_metrics:
```
karl-richter, line 203: #1052

karl-richter:
@noxan I adressed most of our comments:
- renamed `metrics.py` to `utils_metrics.py`
- renamed `collect_metrics` to `metrics_enabled` in `time_net.py`
- renamed `checkpointing` to `checkpointing_enabled` (and similar for metrics, progress_bar etc.)
- fixed non-pythonic return in `_train()` (no more if else for return)

noxan (approved):
LGTM

PR #1040, opened 2022-12-05
Refactored dataloader

karl-richter:
## :microscope: Background

- The dataloader currently uses a for loop in the `__getitem__` function to retrieve a sample from the dataset. Since for a dataset with 10.000 samples that is trained for 100 epochs, this function is called 1 Mio. times, pre-computing the samples should bring a speed-up of the model training.

## :crystal_ball: Key changes

- Pre-compute the samples to only use indexing during retrieval.


noxan (approved):
Great improvement to have it once and not with every iteration 👍

Issue #1017, opened 2022-12-02
Conformal Prediction V1.1 tasks


noxan:
Issue to track the follow up tasks from https://github.com/ourownstory/neural_prophet/pull/802/

- [X] There is quite a lot of code which is currently disabled as comments, we should either finish the implementation or remove the unused comments ([ref](https://github.com/ourownstory/neural_prophet/pull/802#discussion_r1037701222))
- [x] There seems to be an open todo with auto regression enabled and assertion of n_train >= 1 ([ref](https://github.com/ourownstory/neural_prophet/pull/802#discussion_r1037700769))
- [x] The `conformalize_predict` method is currently only commented, do we finish this or remove it again? ([ref](https://github.com/ourownstory/neural_prophet/pull/802/files/fde920ee22a08ed912870843e03e9563a1c0cd0e#r1037699461))

Kevin-Chen0:
Additional tasks:

[] -

noxan:
I believe closing this issue was not intentional

Kevin-Chen0:
Moved the following tasks Conformal Prediction V1.2 https://github.com/ourownstory/neural_prophet/issues/1101:
- Documentation We currently have three ways to work with uncertainty (QR, CP, and CQR): for me it's not really clear when to use which and why? Could either be addressed in some tutorial or if the API would be more similar I might get a better idea of what is the difference and how I switch between them. ([ref](https://github.com/ourownstory/neural_prophet/pull/802#discussion_r1032853973))
- Create conformal prediction method `auto`, which automatically selects "best" conformal prediction method based on the NP params (e.g. quantiles included or not).
- Add `.conformal_test()` method in forecaster.py to wrap around the `.test()`, just like `.conformal_predict()` with `.predict()`.
- Add Continuous Ranked Probability Score (CRPS) as a metric for probabilistic forecast. [Source](https://datumorphism.leima.is/cards/time-series/crps/)

PR #1044, opened 2022-12-06
Uncertainty: Conformal Prediction V2 - add .conformal_predict() and remove .conformalize() in forecaster.py

Kevin-Chen0:
Resolves a task in Issue #1017.

In forecaster.py, add `.conformal_predict()` method that combines the `.conformalize()` and `.predict()`, and will need to input most of the calibration and test sets for split CP. With this implemented, `.conformalize()` will then be removed. Also, the `config_conformal` will no longer be necessary and will be removed.

For future bootstrapped (e.g. Jackknife+ and CV+ based) CP methods, a calibration set will not be needed, which is why calibration_df is set to None as default in conformal_predict().


neuralprophet/forecaster.py
```diff
@@ -3021,15 +3006,17 @@ def _reshape_raw_predictions_to_forecst_df(self, df, predicted, components):
                         df_forecast = pd.concat([df_forecast, yhat_df], axis=1, ignore_index=False)
         return df_forecast
 
-    def conformalize(self, df_cal, alpha, method="naive", plotting_backend="default"):
-        """Apply a given conformal prediction technique to get the uncertainty prediction intervals (or q-hats).
+    def conformal_predict(self, df, alpha, calibration_df=None, method="naive", plotting_backend="default"):
```
noxan, line 3009: @karl-richter I believe you wanted to add a few more parameters to the regular `predict` function - not sure how we deal with multiple variants of the prediction function, we could replicate the additional parameters and pass them through, thoughts?

neuralprophet/forecaster.py
```diff
@@ -3045,21 +3032,22 @@ def conformalize(self, df_cal, alpha, method="naive", plotting_backend="default"
...
-                df_cal, alpha, method, self.config_train.quantiles, plotting_backend
-            )
-
-    # def conformalize_predict(self, df, df_cal, alpha, method="naive"):
-    #     self.conformalize(df_cal, alpha, method)
-    #     df_forecast = self.predict(df)
-    #     return df_forecast
+        q_hats = conformalize(df_cal, alpha, method, self.config_train.quantiles, plotting_backend)
+        # predict
+        df = self.predict(df)
```
noxan, line 3041: we should allow users to pass arguments to the prediction function, also see comment above.

noxan:
Not really sure which is the best API, having an additional setup method or wrapping the predict method - will we have more comparable cases in the future (like other uncertainty prediction modes)? would be great if we can find the most generic approach.

I like the reduction of code 👍 

Added a few minor remarks. Happy to hear your thoughts @Kevin-Chen0, @ourownstory and @karl-richter.

neuralprophet/forecaster.py
Kevin-Chen0: How about pass in **kwargs into `conformal_predict()`, which will then be passed in the `self.predict()`?

neuralprophet/forecaster.py
```diff
@@ -3045,21 +3032,22 @@ def conformalize(self, df_cal, alpha, method="naive", plotting_backend="default"
...
+        df["qhat1"] = q_hats[0]
+        if method == "naive":
+            df["yhat1 - qhat1"] = df["yhat1"] - q_hats[0]
+            df["yhat1 + qhat1"] = df["yhat1"] + q_hats[0]
+        else:  # method == "cqr"
```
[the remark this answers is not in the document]
Kevin-Chen0: You mean throw `ValueError`? If that is the conversion in forecaster.py then sure.

neuralprophet/forecaster.py
Kevin-Chen0: Pass in **kwargs?

Kevin-Chen0:
> Not really sure which is the best API, having an additional setup method or wrapping the predict method - will we have more comparable cases in the future (like other uncertainty prediction modes)? would be great if we can find the most generic approach.
> 
> I like the reduction of code +1
> 
> Added a few minor remarks. Happy to hear your thoughts @Kevin-Chen0, @ourownstory and @karl-richter.

I prefer this PR of wrapping the predict method as `conformal_predict()`, because all of my conformal prediction code sit nicely into `conformal_predict()` w/o intruding into other codes. If I have multiple setup methods like `conformalize()` (as the current setup), then my conformal code will be spread out in forecaster.py, including parts in `predict()`, in `config_conformal`, etc.

That is how with this PR, I can achieve a reduction in the code. Nevertheless, I'm happy to hear other people's thoughts.

neuralprophet/forecaster.py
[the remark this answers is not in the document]
Kevin-Chen0: Implemented.

neuralprophet/forecaster.py
Kevin-Chen0: Implemented.

neuralprophet/forecaster.py
Kevin-Chen0: Implemented.

neuralprophet/forecaster.py
```diff
@@ -3021,13 +3006,17 @@ def _reshape_raw_predictions_to_forecst_df(self, df, predicted, components):
                         df_forecast = pd.concat([df_forecast, yhat_df], axis=1, ignore_index=False)
         return df_forecast
 
-    def conformalize(self, df_cal, alpha, method="naive", plotting_backend="default"):
-        """Apply a given conformal prediction technique to get the uncertainty prediction intervals (or q-hats).
+    def conformal_predict(
+        self, df, calibration_df=None, alpha=0.1, method="naive", plotting_backend="default", **kwargs
```
noxan, line 3010: Not sure myself: Should we better explicitly define the parameters here? If people use the docstrings and code reference to understand this method they won't get far like this. 🤔

neuralprophet/forecaster.py
```diff
@@ -3044,22 +3033,27 @@ def conformalize(self, df_cal, alpha, method="naive", plotting_backend="default"
                     * ``plotly``: Use the plotly backend for plotting
                     * ``matplotlib``: Use matplotlib backend for plotting
                     * ``default`` (default): Use matplotlib backend for plotting
+            kwargs : dict
+                additional predict parameters for test df
         """
-        df_cal = self.predict(df_cal)
+        # conformalize
+        df_cal = self.predict(calibration_df)
```
noxan, line 3040: What does `df_cal` stand for? Could we find a more verbose and clear name instead?

noxan (approved):
LGTM, would be curious to hear your thoughts on the two minor remarks @Kevin-Chen0 

In the long term I am not too happy with having two separate `predict` functions and so different APIs for uncertainty methods - let's discuss this somewhen later.

PR #1045, opened 2022-12-06
Normalize training time metric

karl-richter:
## :microscope: Background

- The training time metric depends on the system performance, but github action workers have varying system specifications.

## :crystal_ball: Key changes

- This pr normalizes the training time based on the currently available system specs.


karl-richter:
This function assumes a linear relationship between training time and CPU performance test (we use the rule of three for calculation). If the relationship is not linear, this might explain why the training times do not always match up.

noxan (approved):

Issue #991, opened 2022-11-24
Windows tests in github actions fail


noxan:
The tests failed on windows multiple occasions now with the following message: `This probably means that Tcl wasn't installed properly.` If you rerun the tests it mostly works, so seems to be a non persistent issue.

[Example: Failed test run](https://github.com/ourownstory/neural_prophet/actions/runs/3537988660/jobs/5938433868)

<img width="1627" alt="Screenshot 2022-11-23 at 10 16 23 PM" src="https://user-images.githubusercontent.com/536225/203708228-2a31a287-8e1c-443c-8e56-645f0b49c221.png">

noxan:
> More than 20 figures have been opened. Figures created through the pyplot interface (`matplotlib.pyplot.figure`) are retained until explicitly closed and may consume too much memory. (To control this warning, see the rcParam `figure.max_open_warning`). Consider using `matplotlib.pyplot.close()`.

noxan:
@karl-richter had the two following resources
1. https://github.com/microsoft/azure-pipelines-tasks/issues/16426
2. https://github.com/orgs/community/discussions/26434

PR #1048, opened 2022-12-07
Test PR for windows tests failing on GitHub due to plotting backend issues

LeonieFreisinger:
#991


LeonieFreisinger:
> @noxan I have investigated the issue. Let me briefly describe you the steps I followed:
> 
> 1. checkout new branch based on your branch `991-windows-tests-in-github-actions-fail` and create new draft PR [Test PR for windows tests failing on GitHub due to plotting backend issues #1048](https://github.com/ourownstory/neural_prophet/pull/1048) to check if windows test fails as well. Note: Locally I am on a python3.8 venv and all tests finish successfully--> Result: Windows test 3.10 fails as well, see image. Further, macos 3.7, macos 3.10 and windows 3.7 tests where aborted.
>    image
> 2. rerun windows 3.10 test online --> result: test finishes successfully.
> 
> so far the same behavior as you observed
> 
> 3. make local changes and push again to trigger online pytests. --> result: all tests finish successfully.
> 4. change local venv to python 3.10.8 (same as used online in the pytest)  and run tests locally.  --> results: all test finish successfully
> 5. push again to trigger online pytests. --> result: all tests finish successfully.
> 
> --> Summary: In my case, the test only fails for the first time (i.e. when creating a new PR) . However, so far could not find its cause.

@noxan found this discussion that might help: https://github.com/actions/runner-images/issues/4931

karl-richter:
I had another look into the error message + the resource you send and the issues they describe seem very plausible. So the issue is that the tcl installation cannot be found (following line). 
```
_tkinter.TclError: Can't find a usable init.tcl in the following directories: 
     {c:\hostedtoolcache\windows\python\3.7.9\x64\tcl\tcl8.6}

c:/hostedtoolcache/windows/python/3.7.9/x64/tcl/tcl8.6/init.tcl: couldn't read file 
```

As they point out here (`https://github.com/pypa/virtualenv/issues/93`), tcl might actually be installed correctly, just in a different location. So the work around they describe above, to simply copy the library from the shared libraries to the location above could be a simple workaround. Otherwise we can maybe set the tcl path to the global location.

Another thing that I just noticed, it says `hostedtoolcache` in the beginning of the path. Could it be that we only have this issue since we cache the installed libraries @noxan ? Maybe thats another direction to solve the root of the cause.

LeonieFreisinger:
not relevant anymore

PR #1067, opened 2022-12-14
[enhancement] Add conditional seasonality as a new feature 

judussoari:
## :microscope: Background

NeuralProphet forces a weekly, daily, and yearly seasonality to the data (if activated). Apart from that, we were able to add custom seasonalities to the model. However, both custom and default seasonalities are applied to the entire dataset, without the possibility to specify a condition for a custom seasonality.

## :crystal_ball: Key changes

`condition_name` is a new input parameter for `m.add_seasonality()`. Using this, the NP model fits and predicts seasonalities according to the condition specified by the user. 
The condition is specified through a new column that is added to the input `df`. The new column should exclusively contain boolean values. `True`, if the condition applies to the row timestamp, and `False` otherwise. The same column(s) must be added to the `future` dataframe, if we want to predict into the unknown future.

The user workflow follows the [example of Facebook Prophet](https://facebook.github.io/prophet/docs/seasonality,_holiday_effects,_and_regressors.html )

See `test_custom_seasons()` for an exmaple: on-season and off-season weekly seasonalities for Football. on-season we see a high during the weekends, while during the week there is no game going on. off-season does not show an increase during the weekend. 

Bildschirmfoto 2022-12-13 um 17 04 15

### DIFF
--- a/neuralprophet/forecaster.py
+++ b/neuralprophet/forecaster.py
@@ -635,7 +635,8 @@ class NeuralProphet:
 
         If condition_name is provided, the dataframe passed to `fit` and
         `predict` should have a column with the specified condition_name
-        containing booleans which decides when to apply seasonality.
+        containing only zeros and ones, deciding when to apply seasonality.
+        Floats between 0 and 1 can be used to apply seasonality partially.
 
         Parameters
         ----------
--- a/neuralprophet/forecaster.py
+++ b/neuralprophet/forecaster.py
@@ -409,6 +409,7 @@ class NeuralProphet:
             weekly_arg=weekly_seasonality,
             daily_arg=daily_seasonality,
             global_local=season_global_local,
+            condition_name=None,
         )
 
         # Events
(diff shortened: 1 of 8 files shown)


neuralprophet/forecaster.py
```diff
@@ -625,13 +626,17 @@ def add_country_holidays(self, country_name, lower_window=0, upper_window=0, reg
...
 
-    def add_seasonality(self, name, period, fourier_order):
+    def add_seasonality(self, name, period, fourier_order, condition_name=None):
         """Add a seasonal component with specified period, number of Fourier components, and regularization.
 
         Increasing the number of Fourier components allows the seasonality to change more quickly
         (at risk of overfitting).
         Note: regularization and mode (additive/multiplicative) are set in the main init.
 
+        If condition_name is provided, the dataframe passed to `fit` and
```
ourownstory, line 636: let's also support  binary values and values from 0 to 1 (that sum up to 1 for each timestamp).

neuralprophet/forecaster.py
```diff
@@ -625,13 +626,18 @@ def add_country_holidays(self, country_name, lower_window=0, upper_window=0, reg
...
+        `predict` should have a column with the specified condition_name
+        containing only zeros and ones, deciding when to apply seasonality.
+        Floats between 0 and 1 can be used to apply seasonality partially.
```
ourownstory, line 639: Can you please add a (code) example - using quarterly changing weekly seasonality - to the docstring?

ourownstory (changes requested):
Could you please also add a quick example to the seasonal tutorial (or a new tutorial notebook if that one feels too long already)?

ourownstory (approved):
Good work!
I think we can merge this now.

Please add in a subsequent PR 
- docstring example
- tutorial inclusion
- some validation of the values in the condition column (range 0-1)

PR #1073, opened 2022-12-17
Uncertainty: Conformal Prediction V1.1 - extend to multiple forecast steps instead of only a single forecast step

Kevin-Chen0:
## :microscope: Background

- Resolves the multiple forecast steps task found in Issue #1017.

## :crystal_ball: Key changes

- Extend to multiple forecast steps instead of only a single forecast step.


Kevin-Chen0:
FYI, I had to comment out `test_PeytonManning`, `test_YosemiteTemps`, and `test_AirPassengers` in test_model_performance.py as they were failing for me.

neuralprophet/conformal.py
```diff
@@ -48,34 +51,50 @@ def predict(self, df: pd.DataFrame, df_cal: pd.DataFrame) -> pd.DataFrame:
...
+                )
+                df[f"yhat{step_number} {quantile_lo}% + qhat{step_number}"] = (
+                    df[f"yhat{step_number} {quantile_lo}%"] + q_hat
+                )
+            else:
+                raise ValueError(
+                    f"Unknown conformal prediction method '{self.method}'. Please input either 'naive' or 'cqr'."
+                )
+            if step_number == 1:
+                # save nonconformity scores of the first timestep
```
ourownstory, line 83: Why are these saved for the first step (only)?

neuralprophet/conformal.py
```diff
@@ -146,8 +170,12 @@ def plot(self, plotting_backend: str):
         """
         method = self.method.upper() if "cqr" in self.method.lower() else self.method.title()
         if plotting_backend == "plotly":
-            fig = plot_nonconformity_scores_plotly(self.noncon_scores, self.alpha, self.q_hat, method)
+            fig = plot_nonconformity_scores_plotly(self.noncon_scores, self.alpha, self.q_hats[0], method)
         elif plotting_backend == "matplotlib":
-            fig = plot_nonconformity_scores(self.noncon_scores, self.alpha, self.q_hat, method)
+            if self.n_forecasts == 1:
+                # includes nonconformity scores of the first timestep
+                fig = plot_nonconformity_scores(self.noncon_scores, self.alpha, self.q_hats[0], method)
```
ourownstory, line 177: Ideally we would have the same behavior for `self.n_forecasts == 1` and other values of `n_forecasts`. 
Maybe we could instead have the special plot that includes the scores be a separate plotting utility call instead of automatically overwriting the standard plotting utility.

ourownstory: Further, there seems to be these two options only in the matplotlib case but not for plotly - However the plot should be as identical as possible independent of the plotting backend. How can we make them return the same kind of plot?

neuralprophet/plot_forecast_matplotlib.py
```diff
@@ -470,8 +470,37 @@ def plot_nonconformity_scores(scores, alpha, q, method):
...
+                * (default) ``naive``: Naive or Absolute Residual
+                * ``cqr``: Conformalized Quantile Regression
+
+    Returns
+    -------
+        matplotlib.pyplot.figure
+            Figure showing the q-values for each timestep
+    """
+    fig, ax = plt.subplots()
+    ax.plot(range(1, len(q_hats) + 1), q_hats)
```
ourownstory, line 500: I am sorry, I am a bit confused about what is being plotted based on the docstring and code.
Aren't the qhats static? 
Or what are we plotting here?

neuralprophet/conformal.py
Kevin-Chen0: The nonconformity scores are saved in order to be inputted into `plot_nonconformity_scores()` , where it plots:
Naive One-Sided Interval Width with q

The nonconformity score is the score (blue line).

This plot is only called when n_forecasts ==1. For models with n_forecasts >1, it will call the `plot_interval_width_per_timestep()`  instead (more details under your 3rd question), which doesn't require nonconformity scores. Therefore, only the first step needs to be saved.

neuralprophet/conformal.py
Kevin-Chen0: Two different plots show whether n_forecasts == 1 or n_forecasts >1. See your 1st and 3rd questions for details.

As for the matplotlib and plotly, both have the `plot_nonconformity_scores()`, but only matplotlib has the `plot_interval_width_per_timestep()`. I'll try to code that for plotly now and before the merge of this PR.

neuralprophet/plot_forecast_matplotlib.py
Kevin-Chen0: This is the `plot_interval_width_per_timestep()` method for models with n_lags > 1 and thus self.n_forecasts >1. You can try with the m3 and m4 models in the [uncertainty_conformal_prediction.ipynb](https://github.com/ourownstory/neural_prophet/blob/main/tutorials/feature-use/uncertainty_conformal_prediction.ipynb). Therefore, instead of plotting the _Naive One-Side Interval Width from q_ (with the noncon scores and q1 horizontal line from just forecast1), this method will instead plot the q values for each of the timesteps:
Naive One-Sdied Interval Width with q
You can see how q linearly increases the further the timestep (until it plateaus around t+12)? This shows the greater uncertainty, the further out the forecast that it requires wider intervals for the same confidence level (e.g., 90% in this example).

neuralprophet/conformal.py
Kevin-Chen0: Just added `plot_interval_width_per_timestep()` for for plotly and modified the `test_plot_conformal_prediction` so the test coverage for that addition is there.

neuralprophet/plot_forecast_matplotlib.py
ourownstory: Thank you for explaining! 
Sorry, I interpreted x-axis to be datetime, not forecast step numbers.
Now it makes sense. Cool how one can see it increase and then plateau!

neuralprophet/conformal.py
ourownstory: Thank you for the explanation!
From a UI-perspective it's not ideal to have the plot change quite so drastically depending on an external factor, but I think it is OK in this case as this is more of a research/diagnostic plot.

neuralprophet/conformal.py
ourownstory: It makes more sense to me now, thank you!
From a UI-perspective it's not ideal to have the plot change quite so drastically depending on an external factor, but I think it is OK in this case as this is more of a research/diagnostic plot.

ourownstory (approved):
Thank you for adding the plotly equivalent! Good work!
Ready to merge.

PR #1078, opened 2022-12-20
Trend modularization

karl-richter:
## Background

- To reduce complexity in the TimeNet, each component should become its own module on the long run.

## Key changes

- This PR contains the minimal changes to migrate the Trend to a separate component.

## Review Checklist


alfonsogarciadecorral:
Hey @karl-richter 

Nice changes. Two things from me:
- Should we have a trend class and then derived classes which represent different models/configuration? In that case should the trend base be an abstract class instead?
- I started to think about which is the most modular way to give the combinations of trend_modl_type  and glocal/local/global. I'm not fully sure, I need a little bit more time. Im going to use the Tuesday 27th to go through it and try to have a proposal then :) 

apart from that it all makes sense

karl-richter:
> Hey @karl-richter
> 
> Nice changes. Two things from me:
> 
> * Should we have a trend class and then derived classes which represent different models/configuration? In that case should the trend base be an abstract class instead?
> * I started to think about which is the most modular way to give the combinations of trend_modl_type  and glocal/local/global. I'm not fully sure, I need a little bit more time. Im going to use the Tuesday 27th to go through it and try to have a proposal then :)
> 
> apart from that it all makes sense

@alfonsogarciadecorral @ourownstory  
As we discussed, I performed the following changes:
- Added a router function to initialize the correct trend based on multiple conditions (`components/router.py`)
- Separated the pure linear trend from the piecewise linear trend
- Added a static trend (only the bias, no growth)
- Added a test for the trend without changepoints

alfonsogarciadecorral (approved):
Great stuff!
I just made a small change as we don't need the `get_trend_deltas` property for the StaticTrend

neuralprophet/components/base.py
```diff
@@ -0,0 +1,35 @@
...
+
+class BaseComponent(nn.Module):
+    def __init__(self, n_forecasts, quantiles, id_list, device):
+        super().__init__()
+        self.n_forecasts = n_forecasts
+        self.quantiles = quantiles
+        self.id_list = id_list
+        self.device = device
+
+    def forward(self, x):
```
ourownstory, line 13: `@abstract`

neuralprophet/components/router.py
```diff
@@ -0,0 +1,69 @@
...
+
+def get_trend(config, id_list, quantiles, num_trends_modelled, n_forecasts, device):
+    """
+    Router for all trend classes.
+
+    Based on the conditions provided, the correct trend class is returned and initialized using the provided args.
+    """
+    if config.growth == "off":
+        # No trend
+        return StaticTrend(
```
ourownstory, line 14: Please complete docstrings.

We could also parse the function args into a dict and then parse(`**`) that into the appropriate trend classes (as all the args seem to be identical and identically named)

neuralprophet/components/base.py
```diff
@@ -0,0 +1,35 @@
...
+        """
+        Needs to be implemented by subclass.
+        """
+        pass
+
+    def new_param(self, dims):
```
ourownstory, line 19: Let's make this a util function?

neuralprophet/components/base.py
ourownstory: alternative to having separate Global/Local/Glocal classes:
introduce additional arg
```
if self.glocal == "global":
    return forward_global()
elif ...
```

neuralprophet/components/trend/trend.py
```diff
@@ -0,0 +1,53 @@
...
+
+        # dimensions  - [no. of quantiles, 1 bias shape]
+        self.bias = self.new_param(
+            dims=[
+                len(self.quantiles),
+            ]
+        )
+
+    @abstractmethod
+    def forward(self, t, meta):
```
ourownstory, line 24: was this already defined in base component?

neuralprophet/components/structure.md
```diff
@@ -0,0 +1,11 @@
+```mermaid
```
ourownstory, line 1: convert to wiki

neuralprophet/components/trend/linear.py
```diff
@@ -0,0 +1,105 @@
+import torch
+import torch.nn as nn
+
+from neuralprophet.components.trend import Trend
+
+
+class LinearTrend(Trend):
```
ourownstory, line 7: evaluate using dataclasses (for all comonents)

ourownstory (approved):
Great Work!
Jut added minor comments, feel free to merge once you are happy with how you addressed them.

PR #1083, opened 2022-12-22
add plotly resampler

LeonieFreisinger:
will close #1076 since that one contains notebooks that should not be merged. 

## :microscope: Background

For **large datasets**, the **plotly** plotting_backend entails **large computation time** (in comparison to matplotlib) for plotting a figure. Additionally, when using an interactive renderer the plotted interactive figure in jupyter notebooks causes a **screen freeze** or **lagged interaction**.
**Goal:** Ideally, we will enable interactive rendering without lagged or freezing interaction.

## :crystal_ball: Key changes
- introduce [plotly-resampler](https://github.com/predict-idlab/plotly-resampler) to **reduze the granuarity** of figures. Plotly-resampler downsamples the data respective to a view.
- Downsampling is ensured for using the `'auto' `mode in all plotly-related `.py `files. --> `register_plotly_resampler(mode='auto')`
- Important note: Input data needs to have a `dtype `--> `list `is not accepted as input data. --> Change `lists `to `np.arrays`

- ([ ] I have added pytests to check whether my feature / fix works.) --> not necessary


LeonieFreisinger:
@ourownstory The plotly resampler is ready to be reviewed. :) I have not changed the default plotting_backend yet, will handle this in a separate PR. 
@karl-richter Would be amazing to get a quick review from you since you are very familiar with plotly. The changes are not too extensive, so the review should be fast. :) I have checked back on the resampler functionality for all plotly-related functions. 

Thank you!

neuralprophet/plot_model_parameters_plotly.py
```diff
@@ -790,7 +792,7 @@ def plot_custom_season(m, comp_name, quantile, multiplicative=False, df_name="__
             fill="none",
         )
     )
-    padded_range = get_dynamic_axis_range(list(range(len(t_i))), type="numeric")
+    padded_range = get_dynamic_axis_range(t_i, type="numeric")
```
karl-richter, line 795: is this code correct? Looks so much simpler than before (if yes, just resolve the comment)

neuralprophet/plot_forecast_plotly.py
```diff
@@ -39,6 +40,7 @@
     "title": dict(font=dict(size=12)),
     "hovermode": "x unified",
 }
+register_plotly_resampler(mode="auto")
```
karl-richter, line 43: Is that all we need to change? Looks super elegant to me!!

neuralprophet/plot_model_parameters_plotly.py
```diff
@@ -99,9 +101,9 @@ def plot_trend_change(m, quantile, plot_name="Trend Change", df_name="__df__"):
     start = data_params["ds"].shift
     scale = data_params["ds"].scale
     time_span_seconds = scale.total_seconds()
-    cp_t = []
+    cp_t = np.array([])
```
karl-richter, line 104: Good that we use numpy here. We should in general be strict in. using numpy arrays instead of lists.

karl-richter (approved):
Looks good! I really like that the re-sampler is more or less a one-line change and does not increase code complexity!

ourownstory:
@karl-richter there seems to be a failure in the metrics code - a plot seems to fail - is this connected to this PR?
As it is not a core test, I think we can address it in a separate PR.

ourownstory:
Great work, thank you @LeonieFreisinger

Issue #970, opened 2022-11-16
Add support for timezones


noxan:
**Is your feature request related to a problem? Please describe.**

- NeuralProphet fails if the `ds` values have a timezone definition

**Describe the solution you'd like**

- Would be great to have NeuralProphet convert the timezone into UTC or unix timestamp internally and not fail.
- Possibly show an `info` logging about the automatic transformation of timezones

**Sample dataset with timezone**
https://www.kaggle.com/datasets/nicholasjhana/energy-consumption-generation-prices-and-weather

**Source reference**
https://github.com/ourownstory/neural_prophet/blob/acf053f61ff2d63e9ab473d22d7fc22489d121e7/neuralprophet/df_utils.py#L442-L443

SaumyaBhushan:
Hi @noxan Can I work on this ?

noxan:
@SaumyaBhushan sure, I've assigned the issue to you. If there are any questions happy to support you :)

SaumyaBhushan:
Thanks @noxan

leoniewgnr:
Hey @SaumyaBhushan, is it correct, that you don't want to work on this anymore? Then I would be happy to work on this.

karl-richter:
Resolved by #1085.

PR #1085, opened 2022-12-23
[Minor] Convert timezones to UTC by default

christymctse:
## :microscope: Background

- Fixing issue #970 

## :crystal_ball: Key changes
- Adding UTC as true in the parameter to convert timezone to UTC timestamp
- Removed warning which stated that NP is not supporting timezone specified dataset


noxan (changes requested):
While testing with the energy dataset I receive the following issue:
```
AttributeError                            Traceback (most recent call last)
/var/folders/h5/r0lsf_xx0dgf0y4_7tyr6zx80000gn/T/ipykernel_55050/2978023471.py in <module>
      4 )
      5 
----> 6 metrics = m.fit(df)
      7 
      8 forecast = m.predict(df)

neural_prophet/neuralprophet/forecaster.py in fit(self, df, freq, validation_df, epochs, batch_size, learning_rate, early_stopping, minimal, metrics, progress, checkpointing, continue_training, num_workers)
...
    504             return PeriodProperties(data, orig)
    505 
--> 506         raise AttributeError("Can only use .dt accessor with datetimelike values")

AttributeError: Can only use .dt accessor with datetimelike values
```

noxan:
I had a brief look into it, seems https://stackoverflow.com/questions/69020835/attributeerror-can-only-use-dt-accessor-with-datetimelike-values-but-cells-con might be helpful and https://pandas.pydata.org/docs/reference/api/pandas.Series.tz_localize.html

noxan:
Would also be great to have the new code in place where the old code was, so that all checks and transformations on the `ds` column are in the same place.

ourownstory:
@christymctse What's the status on this?

christymctse:
> @christymctse What's the status on this?

Will be working on this next week

christymctse:
@noxan @ourownstory, sorry about get back late
 Just fixed the bug and it should be working fine on the energy dataset now

ourownstory:
@noxan Can you please give this another review to check whether it is ready for merge? Thank you!

noxan (approved):
Thanks @christymctse for the adjustments, it works for me now - good job 👍

PR #1100, opened 2023-01-07
[Bug-fix] Enable holidays when using global/glocal modeling

### LINKED ISSUE #1099: [bug] Adding country holidays triggers Error when using global/glocal model
Adding country holidays with `m.add_country_holidays()` triggers an index error when the input df contains multiple time series (i.e. when glocal/global modeling is used). Resetting the index of the incoming time series in `TimeDataset.make_events_features()` resolves the problem.

judussoari:
## :microscope: Background

Fixes #1099
Adding country holidays with m.add_country_holidays() triggers an index error when the input df contains multiple time series (i.e. when glocal/global modeling is used).

## :crystal_ball: Key changes

 Resetting the index of the incoming time series in TimeDataset.make_events_features() resolves the problem.

### DIFF
--- a/neuralprophet/time_dataset.py
+++ b/neuralprophet/time_dataset.py
@@ -508,7 +508,7 @@ def make_events_features(df, config_events: Optional[configure.ConfigEvents] = N
         np.array
             All multiplicative event features (both user specified and country specific)
     """
-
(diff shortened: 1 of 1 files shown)


ourownstory (approved):
LGTM

PR #1105, opened 2023-01-09
[refactor] incorporate typing for time_net.py

Kevin-Chen0:
Resolves Issue #1106.

### DIFF
--- a/neuralprophet/time_net.py
+++ b/neuralprophet/time_net.py
@@ -1,7 +1,7 @@
 import logging
 import math
 from collections import OrderedDict
-from typing import Dict, List, Optional, Tuple, Union
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 import numpy as np
 import pytorch_lightning as pl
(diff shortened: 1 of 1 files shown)


neuralprophet/time_net.py
```diff
@@ -880,7 +880,7 @@ def all_covariates(self, covariates):
                 x = x + self.covariate(lags=covariates[name], name=name)
         return x
 
-    def forward(self, inputs, meta=None):
+    def forward(self, inputs: Dict, meta: Dict = None) -> torch.tensor:
```
noxan, line 883:
```suggestion
    def forward(self, inputs: Dict, meta: Dict = None) -> torch.Tensor:
```

I received several warnings about `torch.tensor` being a function instead of a type. Maybe we better use `torch.Tensor` (upper case letter referring to the class instead of the function).

noxan (changes requested):
LGTM to me overall, but we should double check the `torch.tensor` type, see comment.

neuralprophet/time_net.py
Kevin-Chen0: sg

PR #1109, opened 2023-01-11
[Bug-fix] Refactor warning about global normalization to forecaster.py

### LINKED ISSUE #1053: Global-local warning logged every training batch
The log in denormalize() in time_net.py is logged every batch. When training on a large dataset, this results in thousands of logs.

Could we simply remove it @alfonsogarciadecorral 

Line 1293 in `time_net.py`
```python
if not self.config_normalization.global_normalization:
     log.warning("When Global modeling with local normalization, metrics are displayed in normalized scale.")
else:
     ...
```

judussoari:
## :microscope: Background

Fixes #1053 
This code portion produced a warning on every single training step, which leads to huge, unnecessary log output. 
```
if not self.config_normalization.global_normalization:
     log.warning("When Global modeling with local normalization, metrics are displayed in normalized scale.")
else:
     ...
```

## :crystal_ball: Key changes

Moved the warning to `m.fit()`

### DIFF
--- a/neuralprophet/forecaster.py
+++ b/neuralprophet/forecaster.py
@@ -759,6 +759,9 @@ class NeuralProphet:
             log.warning("Progress plot requires metrics to be enabled. Enabling the default metrics.")
             metrics = metrics.get_metrics(True)
 
+        if not self.config_normalization.global_normalization:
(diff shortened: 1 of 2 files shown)


ourownstory (approved):
Great, thank you!

PR #1110, opened 2023-01-11
Avoidance of duplicate code in regularisation

leoniewgnr:
## :microscope: Background
As mentioned in issue #712, there is duplicate code regarding the event and holiday implementation.

## :crystal_ball: Key changes
Two functions were edited to remove duplicate code:

1. In the regularisation of events and holidays (`def reg_func_events` in `utils.py`) duplicate code was removed and put into a helper function.
2. In construction of the event features (`def make_events_features` in `time_dataset.py`) duplicate code was removed and put into a helper function.

- [ ] I have added pytests to check whether my feature / fix works. (I didn't add additional pytests, as it only was refactoring, but no existing pytest fails)


neuralprophet/time_dataset.py
```diff
@@ -512,43 +512,37 @@ def make_events_features(df, config_events: Optional[configure.ConfigEvents] = N
     additive_events = pd.DataFrame()
     multiplicative_events = pd.DataFrame()
 
+    # helper function to avoid redundant code
+    def create_event_features(event, config):
```
LeonieFreisinger, line 516: The function name might be a little confusing since it is almost the same as `make_events_features()`

LeonieFreisinger (changes requested):

leoniewgnr:
@LeonieFreisinger, thanks for you feedback! I included it in the lates commits. But I don't know why `Linters / isort` and `Linters / black` keep failing. 

- `Linters / isort`: The order of inputs is exactly the same as in main
- `Linters / black`: It tells me that it would reformat` utils.py`, but I let black run before committing and there were no changes in `utils.py`. In addition, it suggests reformatting in lines of code where I did not change anything

neuralprophet/utils.py
```diff
@@ -113,6 +113,29 @@ def reg_func_season(weights):
     return reg_func_abs(weights)
 
 
+def _regularize_weights(weights, reg_lambda):
+    """
+    Helper function for reg_func_events to avoid code duplication
```
LeonieFreisinger, line 118: same here

LeonieFreisinger (approved):
Thanks for cleaning up that part of the code!! I have added only two more comments. The rest is good to go.

PR #1114, opened 2023-01-13
Deactivate plotly resampler for static image export

karl-richter:
## :microscope: Background

- The performance tests started to fail at some point, with the figure export being the cause
- Since neigher plotly nor the image export plugin kaleido have been updated recently, we traced this issue down to the resampler, which seems to be not compatbile with kaleido.

## :crystal_ball: Key changes

- Deactivated the resampler for the pytest where a static figure is exported


LeonieFreisinger (approved):
passt

PR #1118, opened 2023-01-17
[docs] Revamp conformal prediction tutorial material

christymctse:
## :microscope: Background
- Guide users through the conformal prediction feature illustration

## :crystal_ball: Key changes
- Add elaboration on the calibration process
- Remove data exploratory coding
- Reduce to only two base models


ourownstory:
@christymctse Checking in - what is the status on this PR?

christymctse:
@ourownstory, have finished updating the naive part and Krishnan is working on the cqr
Since Kevin just merge the performance metrics ytd, we would try to update the responding session in the tutorial material for the coming week.

leoniewgnr (dismissed):
**General**
Wow, it’s really good now! Well done🙌🏼 I just have some minor changes I would propose:
(just fyi, I tested the notebook in google colab)

**Introduction**
To carry out split conformal prediction, **_the_** dataset has to be split into three distinct sets for training, calibration, and testing respectively.

**Data splitting**
Somehow the shapes do not match
Pasted Graphic

**Base model training**
- It took me really long to train both models (9 minutes, still okay but quite long for a tutorial)
- Just a question: If we have a 90% confidence interval, shouldn’t there lie around 90% of the points in the interval? Because in both plots, it seems like a lot of actual y points lie outside the interval?

**Option 2**
Somehow the column names of the table are weird 
#Aggregate the naive performance metrics for m1

To be honest, I don’t think you need to show the tables separately for naive and qcr. I think its fine to just show the combined table with naive and qcr, because its still a very manageable table and it would shorten it a bit.

christymctse:
Thanks @leoniewgnr 

- Just updated the shapes for data splitting
- Agreeing it takes quite some while to train the model. But since we already have got the output in the material and it would ultimately be posted as a blog at our webpage, we don’t expect a lot of the users would run the notebook.
- Confidence interval is difference from the prediction interval (shown in the graph). Confidence interval was referring to how likely the true mean would lay within the interval after repeated sampling which would not only limited to the observed instances (count as 1 sampling). There’re some interesting discussions around the statical logic behind [here](https://stats.stackexchange.com/questions/26450/why-does-a-95-confidence-interval-ci-not-imply-a-95-chance-of-containing-the), if you’re interested.

Kevin-Chen0 (approved):
LGTM, excellent work @christymctse and @kchander92!

Kevin-Chen0:
Glad you liked it @leoniewgnr! A few things in response to your requested change:

- **Introduction:** It is correct that for split conformal prediction, the dataset has to be split into training, calibration, and test sets.
- **Data splitting:** The split shapes now match as it has been fixed.
- **Base model training:** It takes around 9 minutes for a CPU computer due to the complexity of the `m2` model. This is to clearly illustrate the differences in the conformal prediction results between a weaker model `m1` and a stronger model `m2`. We can simplify the second model in a future PR if the runtime length is still an issue.
- **Base model training:** In an ideal world, 90% confidence interval means mean that 90% of actual future y-values will fall within the forecast. But that is usually not the case, just like how forecasts cannot always precisely predict the future. In fact, the bulk of latest conformal prediction / uncertainty estimation literature is focused on narrowing the *expected* error rate (i.e., `alpha` or `1 - confidence interval`) by users and the *actual* error rate (i.e., `miscoverage rate`) by future OOS data. 
- **Option 2:** 3 layers of columns for the evaluation table is necessary when we need to evaluate for multiple timesteps for multiple CP methods.
  - The top layer is the forecast timestep number, which is `yhat1`, `yhat2`, `yhat3`, ...etc. In this tutorial, we are only evaluating a single timestep, hence only `yhat1`.
  - The middle layer is the name of the CP methods. Currently, we have `naive` and `cqr`, but we will have add more CP methods later on, such as EnbPI.
  - The last layer are the evaluation metrics, including `interval_weight` and `miscoverage_rate`.  
- **Option 2:** We removed the showing of the tables separately for naive and qcr, and only the combined.

PR #1128, opened 2023-01-23
[fix] default value of n_changepoints given correctly in the docs

leoniewgnr:
## :microscope: Background
In the guide hyperparemeter-selection, it says 
'`n_changepoints` is the number of changepoints selected along the series for the trend. The default value for this is 5.' 
But the correct value would be 10, being consistent with the code.

## :crystal_ball: Key changes
Changed the value of 5 to 10 in the guide

- [X] I have added pytests to check whether my feature / fix works.


LeonieFreisinger (approved):
Looks good! Thank you :)

noxan (approved):
LGTM, thanks for updating

PR #1140, opened 2023-01-30
[docs] Readme library purpose

leoniewgnr:
## :microscope: Background
It may be unclear for what tasks NeuralProphet is good to use regarding issue #1136 

## :crystal_ball: Key changes
I added a small section "Purpose" in the README.md to explain the focus of NeuralProphet and what the benefits and disadvantages are.


karl-richter (commented):
Added comments, like it in general!

leoniewgnr:
@karl-richter thanks for your feedback!
This is actually done now, but I still have a call next this week with a user who wants to provide feedback. However, I think we can merge that and might include the user's feedback in form of quotes on the website

ourownstory (approved):
Great work, thank you Leonie!

PR #1149, opened 2023-02-07
Change default plotting backend to plotly-resample

### LINKED ISSUE #948: Switch default plotting backend to plotly
**Status quo**
Currently, matplotlib and plotly are the two available plotting backends.

**Refactoring need**
For future versions, we plan to only support the plotly plotting backend. In the first step the default should be switched from matplotllib to plotly. 

**Update**

**Raised issue** with plotly plotting backend : **slow processing (rendering)** of plotly figures in **jupyter notebook** and **occasionally freezing** notebook. 

> **Investigations**: Tried to reproduce problem with the following settings:

- RAM: 16GB
- renderer mode: `fig.show(renderer="notebook_connected")`
- executed` uncertainty_conformal_prediction.ipynb `  in jupyter lab
  ```
  fig1 = m1.highlight_nth_step_ahead_of_each_forecast(1).plot(
    forecast1[-cutoff:], figsize=figsize, plotting_backend=plotting_backend
  )
results: execution time 0.11 seconds and RAM used 14.65 GB (that means could not reproduce the issue with my environment setting) 


further infos on plotly renderers:  https://plotly.com/python/renderers/

### LINKED ISSUE #1127: Rethink registering of plotly resampler
Limit the registration of plotly resampler to functions where it neds to be registered instead of registering full files. This prevents from unintended registration. Think about introducing decorators.

LeonieFreisinger:
close #948 
close #1127 
## :microscope: Background
So far, we have 2 plotting backends in use `plotly` and `mayplotlib`. `plotly` offers an interactive experience, which is nice. However, a seamless interactive experience with `plotly` on large data is not guaranteed. That's why we introduce a resample mode for `plotly` resulting in a third backed `plotly-resample`. The resampling is only supported for interactive environments, i.e. notebooks, expect from google colab. The `plotly-resample` backend will be the default mode. 

## :crystal_ball: Key changes
- Introduce thried plotting backend `plotly-resample`
- Unlock `plotly-resample` for jupyterlab and vscode environments
- register plotly-resampler for individual plotting function if `plotly-resample` is unlocked
- unregitser plotly-resampler in case `plotly-resample` not unlocked
- adapt plotly tutorial

- [ ] Info: I have not added the `plotly-resample` backend to the pytests, since the code ist mostly the same as with `plotly` and it would higher our number of test runs by >10


neuralprophet/plot_utils.py
```diff
@@ -18,6 +18,20 @@ def log_warning_deprecation_plotly(plotting_backend):
...
+def log_warning_colab_resampler():
+    log.warning(
+        "Warning: 'plotly-resample' plotting backend not supported for google colab environment. "
+        "Plotting backend automatically switched to 'plotly' without resampling "
+    )
+
+
+def log_warning_static_env_resampler():
+    log.warning(
+        "Warning: 'plotly-resample' plotting backend not supported for static environments. "
```
LeonieFreisinger, line 30: this

neuralprophet/conformal.py
```diff
@@ -158,35 +159,60 @@ def _get_q_hat(self, df_cal: pd.DataFrame, noncon_scores: np.ndarray) -> float:
...
         Parameters
         ----------
             plotting_backend : str
                 specifies the plotting backend for the nonconformity scores plot, if any
 
-                Options
-                    * ``matplotlib``: Use matplotlib backend for plotting
-                    * ``plotly``: Use the plotly backend for plotting
+                                Options
+                * ``plotly-resample``: Use the plotly backend for plotting in resample mode. This mode uses the
```
LeonieFreisinger, line 171: plotly-resample with the comment that you might not get any error message.
plotly-auto

neuralprophet/forecaster.py
```diff
@@ -1687,6 +1695,18 @@ def plot(
                 figsize=figsize,
                 highlight_forecast=forecast_in_focus,
             )
+        else:  # default plotting backend
```
LeonieFreisinger, line 1698: if 'plotly' in plotting_backend
below: resample_active=plotting_backend == "plotly-resample"

neuralprophet/forecaster.py
```diff
@@ -1604,7 +1609,10 @@ def plot(
                 optional, overwrites the default plotting backend.
 
                 Options
-                * ``plotly``: Use plotly for plotting
+                * ``plotly-resample``: Use the plotly backend for plotting in resample mode. This mode uses the
+                    plotly-resampler package to accelerate visualizing large data by resampling it. Only supported for
```
LeonieFreisinger, line 1613: should we call it identical to the package?

neuralprophet/forecaster.py
```diff
@@ -1664,10 +1672,9 @@ def plot(
         plotting_backend = (
```
LeonieFreisinger, line 1672: if plotting backend is not given by user here (set to default/auto):
- check if it was previously set (if so, use that)
- if not, call function auto_set_plotting_backend(), get one of 3 backends, depending on env. 

if plotting backend is given by user (in a plot function), use, but only for this call (self.plotting_backend)

if user calls set_plotting_backend():
- if default/auto, call function auto_set_plotting_backend(), get one of 3 backends, depending on env. Then write this value to self.plotting_backend
-else, if one of the 3 backends,  writes value to self.plotting_backend

LeonieFreisinger:
@ourownstory Would be great if you could give the plotly resampler a final review. Thank you! Short remark: black and pyright currently seem to fail on the main branch.

neuralprophet/conformal.py
```diff
@@ -158,7 +159,30 @@ def _get_q_hat(self, df_cal: pd.DataFrame, noncon_scores: np.ndarray) -> float:
 
         return q_hat
 
-    def plot(self, plotting_backend: str):
+    def set_plotting_backend(self, plotting_backend):
```
ourownstory, line 162: Do we need this method here, or can we call a util instead?

neuralprophet/forecaster.py
```diff
@@ -1547,14 +1547,19 @@ def set_plotting_backend(self, plotting_backend):
...
+                * (default) ``plotly-resample``: Use the plotly backend for plotting in resample mode. This mode uses the
+                    plotly-resampler package to accelerate visualizing large data by resampling it. Only supported for
+                    jupyterlab notebooks and vscode notebooks.
                 * ``plotly``: Use the plotly backend for plotting
-                * (default) ``matplotlib``: use matplotlib for plotting
+                * ``matplotlib``: use matplotlib for plotting
         """
-        if plotting_backend in ["plotly", "matplotlib"]:
+        if plotting_backend in ["plotly-auto", "plotly", "matplotlib", "plotly-resampler"]:
             self.plotting_backend = plotting_backend
```
ourownstory, line 1557: Are you intentionally allowing to set "plotly-auto" as a backend?
I would expect you to call `auto_set_plotting_backend` and save the returned backend.

neuralprophet/forecaster.py
```diff
@@ -1604,9 +1609,14 @@ def plot(
                 optional, overwrites the default plotting backend.
 
                 Options
-                * ``plotly``: Use plotly for plotting
+                * ``plotly-resampler``: Use the plotly backend for plotting in resample mode. This mode uses the
+                    plotly-resampler package to accelerate visualizing large data by resampling it. For some
+                    environments (colab, pycharm interpreter)plotly-resampler might not properly vizualise the figures.
+                    In this case, consider switching to 'plotly-auto'.
```
ourownstory, line 1615: To be more clear, I would formulate the last line:
`In this case, consider switching to 'plotly' or 'matplotlib'`

neuralprophet/forecaster.py
```diff
@@ -1660,22 +1670,27 @@ def plot(
...
-            plotting_backend
-            if plotting_backend != "default"
-            else (self.plotting_backend if hasattr(self, "plotting_backend") else "matplotlib")
+            auto_set_plotting_backend(plotting_backend)
+            if plotting_backend is not None
+            else (
+                auto_set_plotting_backend(self.plotting_backend)
+                if hasattr(self, "plotting_backend")
+                else auto_set_plotting_backend("plotly-auto")
+            )
```
ourownstory, line 1681: Suggestion: For ease of reading, you could use a more traditional if-else format.

neuralprophet/forecaster.py
```diff
@@ -1824,22 +1844,27 @@ def plot_latest_forecast(
             fcst, self.config_train.quantiles, n_last=1 + include_previous_forecasts
         )
 
-        # Check whether the default plotting backend is overwritten
+        # Check whether a local or global plotting backend is set.
         plotting_backend = (
-            plotting_backend
-            if plotting_backend != "default"
-            else (self.plotting_backend if hasattr(self, "plotting_backend") else "matplotlib")
+            auto_set_plotting_backend(plotting_backend)
```
ourownstory, line 1849: Suggestion: For ease of reading, you could use a more traditional if-else format.

neuralprophet/forecaster.py
```diff
@@ -1824,22 +1844,27 @@ def plot_latest_forecast(
...
+            if plotting_backend is not None
+            else (
+                auto_set_plotting_backend(self.plotting_backend)
+                if hasattr(self, "plotting_backend")
+                else auto_set_plotting_backend("plotly-auto")
+            )
         )
         log_warning_deprecation_plotly(plotting_backend)
-        if plotting_backend == "plotly":
+        if plotting_backend.find("plotly") == 0:
```
ourownstory, line 1858: Suggestion: use more pythonic notation:
`if 'plotly' in plotting_backend.lower():`

neuralprophet/forecaster.py
```diff
@@ -1985,22 +2015,27 @@ def plot_components(
...
-            plotting_backend
-            if plotting_backend != "default"
-            else (self.plotting_backend if hasattr(self, "plotting_backend") else "matplotlib")
+            auto_set_plotting_backend(plotting_backend)
+            if plotting_backend is not None
+            else (
+                auto_set_plotting_backend(self.plotting_backend)
+                if hasattr(self, "plotting_backend")
+                else auto_set_plotting_backend("plotly-auto")
+            )
```
ourownstory, line 2026: Suggestion: For ease of reading, you could use a more traditional if-else format.
Also, this block seems repeated. maybe part of it can be simplified of part of the auto-setter?

neuralprophet/forecaster.py
```diff
@@ -2135,14 +2175,18 @@ def plot_parameters(
...
```
ourownstory, line 2186: see above

neuralprophet/forecaster.py
```diff
@@ -2135,14 +2175,18 @@ def plot_parameters(
...
         )
         log_warning_deprecation_plotly(plotting_backend)
-        if plotting_backend == "plotly":
+        if plotting_backend.find("plotly") == 0:
```
ourownstory, line 2189: see above

neuralprophet/forecaster.py
```diff
@@ -1985,22 +2015,27 @@ def plot_components(
...
+            else (
+                auto_set_plotting_backend(self.plotting_backend)
+                if hasattr(self, "plotting_backend")
+                else auto_set_plotting_backend("plotly-auto")
+            )
         )
         log_warning_deprecation_plotly(plotting_backend)
 
-        if plotting_backend == "plotly":
+        if plotting_backend.find("plotly") == 0:
```
ourownstory, line 2030: see above

neuralprophet/plot_utils.py
```diff
@@ -538,3 +552,72 @@ def get_valid_configuration(  # move to utils
             "components_list": plot_components,
         }
     return valid_configuration
+
+
+def validate_current_env():
```
ourownstory, line 557: This could use a clearer name - what is the env valid for?

neuralprophet/plot_utils.py
```diff
@@ -538,3 +552,72 @@ def get_valid_configuration(  # move to utils
...
+    Returns
+    -------
+    str
+        The new plotting backend.
+    """
+    if plotting_backend_original == "plotly-auto":
+        plotting_backend_new = "plotly-resample" if validate_current_env() else "plotly"
+    else:
+        plotting_backend_new = plotting_backend_original
+    return plotting_backend_new
```
ourownstory, line 623: I think you can use just `plotting_backend` as variable here - that will also shorten you code
```
if plotting_backend == "plotly-auto":
        plotting_backend = "plotly-resample" if validate_current_env() else "plotly"
    return plotting_backend
```

Further suggestion:
This only auto-sets the plotly backend.
Thus it could only be called if plotting backend is "plotly-auto" and not receive any input, and be called
`auto_set_plotly_backend()`

ourownstory (dismissed):
Looks good, thank you!

I left some minor suggestions for coding practices.

Further I did wonder, if we could avoid using both `None` and `plotly-auto` as user inputs. 
The way you have structured it, the most straightforward way would be to do away with  `plotly-auto` alltogeher. 
Then, if `plotting_backend` is `None`, it would `auto_set_plotly_backend()` 
if a user sets a value, it must be one of the three actual backends.
if they explicitly use 'plotly-resample' in a non-supported env, we would just throw an error stating to switch to plotly or matplotlib. 
This would de-complexify the setting process a bit and be as user-friendly due to reduced ambiguity IMO.

Finally, let us set a single test that uses resampler, just catch major issues.

neuralprophet/conformal.py
LeonieFreisinger: Agree! Would move that into a new PR related to the general refactoring of the structure.

LeonieFreisinger:
@ourownstory Thanks for your detailed feedback! I tried to address your suggestions as follows:

- ✅ avoid `plotly-auto` use `None `instead, then `auto_set_plotly_backend()`
    - ✅ exclude `plotly-auto` from `set_plotting_backend`()
    - ✅ restructure `auto_set_plotly_backend`()
- ✅ allow user to only set one of the actual backends
- ✅ single test that uses the resampler backend
- ✅ simplified if-else statement for backend assignment
- replace `if plotting_backend.find("plotly") == 0:` with `if 'plotly' in plotting_backend.lower():`
- ✅ rename `validate_current_env()`  to `validate_current_env_for_resampler()`
    - ✅ add `auto `flag to `validate_current_env_for_resampler()` to trigger different warning
    - ✅ for `plotting_backend == ‘plotly-resampler’` add one warning that goes in line with auto switiching the backend and another one that recommends switching to “plotly” or “matplotlib.
    
Feel free to have another look at it.

neuralprophet/conformal.py
```diff
@@ -167,29 +191,44 @@ def plot(self, plotting_backend: str):
                 specifies the plotting backend for the nonconformity scores plot, if any
 
                 Options
-                    * ``matplotlib``: Use matplotlib backend for plotting
-                    * ``plotly``: Use the plotly backend for plotting
+                * ``plotly-resampler``: Use the plotly backend for plotting in resample mode. This mode uses the
+                    plotly-resampler package to accelerate visualizing large data by resampling it. For some
+                    environments (colab, pycharm interpreter)plotly-resampler might not properly vizualise the figures.
```
noxan, line 196:
```suggestion
                    environments (colab, pycharm interpreter) plotly-resampler might not properly vizualise the figures.
```

neuralprophet/conformal.py
```diff
@@ -167,29 +191,44 @@ def plot(self, plotting_backend: str):
...
         """
         method = self.method.upper() if "cqr" in self.method.lower() else self.method.title()
-        if plotting_backend == "plotly":
+        # Check whether a local or global plotting backend is set.
+        plotting_backend = auto_set_plotting_backend(plotting_backend)
+        if hasattr(self, "plotting_backend"):
+            plotting_backend = auto_set_plotting_backend(self.plotting_backend)
+
+        log_warning_deprecation_plotly(plotting_backend)
+        if "plotly" in plotting_backend.lower():
```
noxan, line 212: minor: might make more sense to ensure `plotting_backend` is always lowercase, otherwise it might cause problems on further checks

noxan, line 212: Also `startswith()` might be the more accurate choice over `in`
```suggestion
        if plotting_backend.lower().startswith("plotly"):
```

neuralprophet/plot_utils.py
```diff
@@ -542,3 +557,84 @@ def get_valid_configuration(  # move to utils
...
+        if "IPKernelApp" not in get_ipython().config:  # pragma: no cover
+            return False
+    except ImportError:
+        return False
+    except AttributeError:
+        return False
+    return True
+
+
+def auto_set_plotting_backend(plotting_backend):
```
noxan, line 622: Bit confused by the function name, this does not `set` anything, it rather `detects` or `finds` the right plotting backend or?

neuralprophet/plot_utils.py
```diff
@@ -542,3 +557,84 @@ def get_valid_configuration(  # move to utils
...
+        The plotting backend to use.
+
+    Returns
+    -------
+    str
+        The new plotting backend.
+    """
+    if plotting_backend is None:
+        plotting_backend = "plotly-resampler" if validate_current_env_for_resampler(auto=True) else "plotly"
+    if plotting_backend == "plotly-resampler":
```
noxan, line 638:
```suggestion
    elif plotting_backend == "plotly-resampler":
```
Does not make sense to run `validate_current_env_for_resampler` again if we just used it with `auto=True` or?

neuralprophet/plot_utils.py
```diff
@@ -542,3 +557,84 @@ def get_valid_configuration(  # move to utils
...
+    plotting_backend: str
+        The plotting backend to use.
+
+    Returns
+    -------
+    str
+        The new plotting backend.
+    """
+    if plotting_backend is None:
+        plotting_backend = "plotly-resampler" if validate_current_env_for_resampler(auto=True) else "plotly"
```
noxan, line 637: Should do the same, yet easier to read
```suggestion
        if validate_current_env_for_resampler(auto=True):
            return "plotly-resampler"
        else:
            return "plotly"
```

neuralprophet/forecaster.py
```diff
@@ -1858,22 +1873,22 @@ def plot_latest_forecast(
...
-        # Check whether the default plotting backend is overwritten
-        plotting_backend = (
-            plotting_backend
-            if plotting_backend != "default"
-            else (self.plotting_backend if hasattr(self, "plotting_backend") else "matplotlib")
-        )
+        # Check whether a local or global plotting backend is set.
+        plotting_backend = auto_set_plotting_backend(plotting_backend)
+        if hasattr(self, "plotting_backend"):
+            plotting_backend = auto_set_plotting_backend(self.plotting_backend)
```
noxan, line 1879: Why do you call `auto_set_plotting_backend` a second time here? For me this reads as the local parameter would be overridden by the global configuration (which is not what we want I assume)?

noxan (changes requested):
Overall it looks good to me, a few minor suggestions - please check the global/local setting for plotting backend, it seems to me it would override the explicit local with the global, maybe I missed something though.

neuralprophet/forecaster.py
LeonieFreisinger: @noxan You are absolutely right! What do you think about the following:
```
plotting_backend = auto_set_plotting_backend(plotting_backend)
if hasattr(self, "plotting_backend") and plotting_backend is None:
     plotting_backend = auto_set_plotting_backend(self.plotting_backend)
```

neuralprophet/conformal.py
```diff
@@ -167,29 +168,45 @@ def plot(self, plotting_backend: str):
...
+                    environments.
 
         """
         method = self.method.upper() if "cqr" in self.method.lower() else self.method.title()
-        if plotting_backend == "plotly":
+        # Check whether a local or global plotting backend is set.
+        if hasattr(self, "plotting_backend") and plotting_backend is None:
+            plotting_backend = select_plotting_backend(self.plotting_backend)
+        else:
+            plotting_backend = select_plotting_backend(plotting_backend)
```
noxan, line 187: What would happen if both conditions are false? Meaning no `self.plotting_backend` and no `plotting_backend` parameter? Then no `select_plotting_backend` is called 🤔

neuralprophet/forecaster.py
```diff
@@ -1858,22 +1874,23 @@ def plot_latest_forecast(
...
-        plotting_backend = (
-            plotting_backend
-            if plotting_backend != "default"
-            else (self.plotting_backend if hasattr(self, "plotting_backend") else "matplotlib")
-        )
+        # Check whether a local or global plotting backend is set.
+        if hasattr(self, "plotting_backend") and plotting_backend is None:
+            plotting_backend = select_plotting_backend(self.plotting_backend)
+        else:
+            plotting_backend = select_plotting_backend(plotting_backend)
```
noxan, line 1881: Minor improvement idea: we have these lines of code several time, how about extracting it into a method?

neuralprophet/forecaster.py
```diff
@@ -1858,22 +1874,23 @@ def plot_latest_forecast(
...
+
         log_warning_deprecation_plotly(plotting_backend)
-        if plotting_backend == "plotly":
+        if plotting_backend.startswith("plotly"):
             return plot_plotly(
                 fcst=fcst,
                 quantiles=self.config_train.quantiles,
-                xlabel=xlabel,
                 ylabel=ylabel,
+                xlabel=xlabel,
```
noxan, line 1889: Any reason to change the sequence here or by mistake?

neuralprophet/conformal.py
LeonieFreisinger: @noxan Exactly, then `select_plotting_backend(None) `would be called. Which selects a plotting_backend based on the input which is None

noxan: True, messed up some boolean logic in my head 👍

noxan (approved):

PR #1167, opened 2023-02-22
Tackle bottlenecks

### LINKED ISSUE #15: Measure computation times, Find bottlenecks and speed up.

### LINKED ISSUE #848: NeuralProphet with numba
TLDR: Has anybody used numba to speed up neuralprophet? And other pytorch based code ? If yes, does it work?

My concern is that since I added lags and regressors, the neural prophet compute got VERY slow and it takes 10 hours to loop over 700 products on which I want to do forecasting on.

Any help is great!

Thanks!

### LINKED ISSUE #1057: [bug] Computation bottleneck for large datasets with multiple time series/IDs
For my dataset (2 million entries, >30 different IDs), we currently run into a significant bottleneck that is caused by something prior to model training. Before model training starts, it somehow processes the data for >2mins. 

Edit: my first suspicion is related to the fourier terms of seasonality. increasing the number of fourier terms for x`_seasonality` leads to a significant increase in overhead time

judussoari:
## :microscope: Background

- Currently, NP execution time scales linearly with # of time series in the (panel) dataset. The goal is to make computation more efficient, and save time
- Closes #1057 
- Closes #848 
- Closes #15 

## :crystal_ball: Key changes
- [ ]  we always compute global and local normalization (both!) even if we do global normalization 
- [ ] there is something, somewhere, in the preprocessing that takes forever to initialize before the training process starts
- [ ]  Multiprocessing for dataloader, should work bc not trainable
- [x] `drop_nan_after_init` is only called in training, no longer in training and prediction (as long as no samples are dropped)
- [x] `.astype()` function is very slow - get rid of it 
- [x] Rethink `df_forecast` output: instead of target-wise, let's switch to origin-wise (raw format) and only reshape for plotting
- [x]  for large datasets: pd.concat() in m.predict() is incredibly inefficient.
- [ ] Use Ray and Anycale to accelerate computation https://speakerdeck.com/anyscale/faster-time-series-forecasting-using-ray-and-anyscale?slide=10
- [x]  `for ID in df.groupby('ID')` must be vectorized (e.g. using `numba`) in `m.predict_raw` and potentially in `m.fit`
- [ ] `init_after_tabularized()` and `drop_nan_after_init()` is slow, can we accelerate it?
- [ ] pywarning regarding df_forecast[yhat] in _reshape function
- [ ] Preprocessing -- return df to user? preprocessing function and move everything out of fit and predict?
- [ ] Understand `num_workers=` param. Apparently, setting it to the maximum no. of cores does not give the best speed?!
- [ ] float32 instead of float64 brings benefit accordding to Darts: https://unit8co.github.io/darts/userguide/torch_forecasting_models.html#performance-recommendations
- [ ] TimeNet covariates are more time consuming than AR net. They are called one after another. Vectorize? 
- [ ] Potentially many more bottlenecks

## Abandoned
- [x] TimeDatasets are very inefficient --> FastTensor Dataloader? https://discuss.pytorch.org/t/dataloader-much-slower-than-manual-batching/27014/5 --> no speed up, do we have real tabularized data?


leoniewgnr:
Will be completed in V2, moved to https://github.com/ourownstory/neural_prophet/wiki/Abandoned-Pull-Requests-and-Issues

PR #935, opened 2022-11-05
Issue #252: Add get/set params

anastasiia-tovstolis:
add 4 new functions: get/set params for neural_prophet and neural_prophet.model

PR #1175, opened 2023-02-27
[minor] Add get function for parameters

### LINKED ISSUE #252: add model.get_params and set_params
similar to scikit model.get_params

for scenario testing or for parameter-based save / load functions

noxan:
Replacement for #935, rebasing the branch and work from @anastasiia-tovstolis

Close #252


neuralprophet/forecaster.py
```diff
@@ -340,6 +340,10 @@ def __init__(
         accelerator: Optional[str] = None,
         trainer_config: dict = {},
     ):
+        kwargs = locals()
+        self.config = kwargs
+        self.config.pop("self")
```
karl-richter, line 345: What exactly does this line do?

karl-richter (approved):

karl-richter:
@noxan please check the pyright test

neuralprophet/forecaster.py
noxan: it removes the `self` reference from the locals() call, so we do only keep the provided arguments

neuralprophet/forecaster.py
```diff
@@ -340,6 +340,10 @@ def __init__(
         accelerator: Optional[str] = None,
         trainer_config: dict = {},
     ):
+        kwargs = locals()
+        self.config = kwargs
```
noxan, line 344:
```suggestion
        self.config = locals()
```

neuralprophet/forecaster.py
noxan: `locals()` returns all locally defined variables (in this case all function parameters)

noxan:
@karl-richter thanks for the notice, i'll address the pyright warnings elsewhere as they came from another PR

PR #1177, opened 2023-02-28
[fix] Fix pyright warnings on wrong types

noxan:


karl-richter (approved):

karl-richter (changes requested):
The tests fail for some reason.

neuralprophet/time_net.py
```diff
@@ -1,7 +1,8 @@
 import logging
 import math
+import np_types
```
karl-richter, line 3: this import does not work

karl-richter, line 3:
```suggestion
from neuralprophet import np_types
```

karl-richter (approved):

Issue #982, opened 2022-11-22
Review our logging setup


noxan:
1. Should we change the default log level to `warning`?
2. There are quite a lot of logging messages shown by default: Which of those `info` log messages should rather `debug` and which `warnings`?

@ourownstory Happy to follow up on our discussion and include other people here :)

noxan:
Duplicate log entries for `m.predict()`
```
INFO - (NP.df_utils._infer_frequency) - Major frequency MS corresponds to 90.972% of the data.
INFO - (NP.df_utils._infer_frequency) - Defined frequency is equal to major frequency - MS
INFO - (NP.df_utils.return_df_in_original_format) - Returning df with no ID column
INFO - (NP.df_utils._infer_frequency) - Major frequency MS corresponds to 90.909% of the data.
INFO - (NP.df_utils._infer_frequency) - Defined frequency is equal to major frequency - MS
INFO - (NP.df_utils._infer_frequency) - Major frequency MS corresponds to 90.909% of the data.
INFO - (NP.df_utils._infer_frequency) - Defined frequency is equal to major frequency - MS
INFO - (NP.df_utils.return_df_in_original_format) - Returning df with no ID column
```

leoniewgnr:
PR will be closed for V2 release, but ideas and feature request are saved and might be realized in V2.

PR #1179, opened 2023-02-28
[refactor] Refactor the prep_or_copy_df method

noxan:
Nothing changed, just more flat control flow and hopefully easier to read. Now also with typing :)

### DIFF
--- a/neuralprophet/df_utils.py
+++ b/neuralprophet/df_utils.py
@@ -57,9 +57,9 @@ def prep_or_copy_df(df: pd.DataFrame):
 
     # Create a list of unique ID values
     unique_id_values: list[str] = df_copy["ID"].unique().tolist()
-    has_unique_id_column = len(unique_id_values) == 1
     # Check if there is only one unique ID value
-    if has_unique_id_column:
+    df_has_single_time_series = len(unique_id_values) == 1
+    if df_has_single_time_series:
         log.debug("Provided DataFrame (df) has an ID column and contains a single time series.")
         return df_copy, df_has_id_column, True, unique_id_values
 
--- a/neuralprophet/df_utils.py
+++ b/neuralprophet/df_utils.py
@@ -39,8 +39,6 @@ def prep_or_copy_df(df: pd.DataFrame) -> tuple[pd.DataFrame, bool, bool, list[st
         list
             list of IDs
     """
-    if df is None:
-        raise ValueError("Provided DataFrame (df) must not be None.")
     if not isinstance(df, pd.DataFrame):
         raise ValueError("Provided DataFrame (df) must be of pd.DataFrame type.")
 
--- a/neuralprophet/df_utils.py
+++ b/neuralprophet/df_utils.py
@@ -22,7 +22,7 @@ class ShiftScale:
     scale: float = 1.0
 
 
-def prep_or_copy_df(df):
+def prep_or_copy_df(df: pd.DataFrame):
     """Copy df if it contains the ID column. Creates ID column with '__df__' if it is a df with a single time series.
     Parameters
     ----------
--- a/neuralprophet/df_utils.py
+++ b/neuralprophet/df_utils.py
@@ -60,10 +60,8 @@ def prep_or_copy_df(df: pd.DataFrame):
     # Check if there is only one unique ID value
     df_has_single_time_series = len(unique_id_values) == 1
 
-    if df_has_single_time_series:
-        log.debug("Provided DataFrame (df) has an ID column and contains a single time series.")
-    else:
-        log.debug("Provided DataFrame (df) has an ID column and contains multiple time series.")
+    single_or_multiple_message = "a single" if df_has_single_time_series else "multiple"
+    log.debug(f"Provided DataFrame (df) has an ID column and contains {single_or_multiple_message} time series.")
 
     return df_copy, df_has_id_column, df_has_single_time_series, unique_id_values
 
--- a/neuralprophet/df_utils.py
+++ b/neuralprophet/df_utils.py
@@ -57,8 +57,9 @@ def prep_or_copy_df(df: pd.DataFrame):
 
     # Create a list of unique ID values
     unique_id_values: list[str] = df_copy["ID"].unique().tolist()
+    has_unique_id_column = len(unique_id_values) == 1
     # Check if there is only one unique ID value
-    if len(unique_id_values) == 1:
+    if has_unique_id_column:
         log.debug("Provided DataFrame (df) has an ID column and contains a single time series.")
         return df_copy, df_has_id_column, True, unique_id_values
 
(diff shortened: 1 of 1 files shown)


neuralprophet/df_utils.py
```diff
@@ -36,31 +36,34 @@ def prep_or_copy_df(df):
...
-            new_df["ID"] = "__df__"
-            log.debug("Received df with single time series")
-    elif df is None:
-        raise ValueError("df is None")
-    else:
-        raise ValueError("Please, insert valid df type (pd.DataFrame)")
+    if df is None:
+        raise ValueError("Provided DataFrame (df) must not be None.")
+    if not isinstance(df, pd.DataFrame):
+        raise ValueError("Provided DataFrame (df) must be of pd.DataFrame type.")
```
noxan, line 45: Does it make a difference if people added a `None` value or the wrong type, we could combine those and say it must be of type `pd.DataFrame`.

neuralprophet/df_utils.py
```diff
@@ -36,31 +36,34 @@ def prep_or_copy_df(df):
...
+    # Create a copy of the dataframe
+    df_copy = df.copy(deep=True)
+
+    df_has_id_column = "ID" in df_copy.columns
+
+    # If there is no ID column, then add one with a single value
+    if not df_has_id_column:
+        log.debug("Provided DataFrame (df) contains a single time series.")
+        df_copy["ID"] = "__df__"
+        return df_copy, df_has_id_column, True, ["__df__"]
```
noxan, line 56: The return here could also be skipped, would possible improve readability, but we kind of do unnecessary calculations later on.

neuralprophet/df_utils.py
```diff
@@ -36,31 +36,34 @@ def prep_or_copy_df(df):
...
+        raise ValueError("Provided DataFrame (df) must be of pd.DataFrame type.")
+
+    # Create a copy of the dataframe
+    df_copy = df.copy(deep=True)
+
+    df_has_id_column = "ID" in df_copy.columns
+
+    # If there is no ID column, then add one with a single value
+    if not df_has_id_column:
+        log.debug("Provided DataFrame (df) contains a single time series.")
```
noxan, line 54: We could mention the originally missing ID column or the fact that one was automatically added?

neuralprophet/df_utils.py
noxan: `isinstance(None, pd.DataFrame) = None`

noxan, line 45:
```suggestion
    if not isinstance(df, pd.DataFrame):
        raise ValueError("Provided DataFrame (df) must be of pd.DataFrame type.")
```

neuralprophet/df_utils.py
karl-richter: Sounds good.

neuralprophet/df_utils.py
```diff
@@ -36,31 +36,32 @@ def prep_or_copy_df(df):
...
-            log.debug("Received df with single time series")
-    elif df is None:
-        raise ValueError("df is None")
-    else:
-        raise ValueError("Please, insert valid df type (pd.DataFrame)")
+    if not isinstance(df, pd.DataFrame):
+        raise ValueError("Provided DataFrame (df) must be of pd.DataFrame type.")
+
+    # Create a copy of the dataframe
+    df_copy = df.copy(deep=True)
```
karl-richter, line 46: do we need the deep=True arg here? Maybe we should try if this also works without, I could imagine that this adds quite some overhead for large datasets.

karl-richter (approved):
Looks good! Maybe quickly check if deep=True is really needed in the copy statement, otherwise ready to go from my side.

neuralprophet/df_utils.py
```diff
@@ -36,31 +36,32 @@ def prep_or_copy_df(df):
...
-                log.debug("Received df with single time series")
-        else:
-            new_df["ID"] = "__df__"
-            log.debug("Received df with single time series")
-    elif df is None:
-        raise ValueError("df is None")
-    else:
-        raise ValueError("Please, insert valid df type (pd.DataFrame)")
+    if not isinstance(df, pd.DataFrame):
+        raise ValueError("Provided DataFrame (df) must be of pd.DataFrame type.")
```
LeonieFreisinger, line 43: @noxan Is it common to wrap the variable name in round brackets, in this case (df)?

neuralprophet/df_utils.py
```diff
@@ -36,31 +36,32 @@ def prep_or_copy_df(df):
...
+
+    # Create a copy of the dataframe
+    df_copy = df.copy(deep=True)
+
+    df_has_id_column = "ID" in df_copy.columns
+
+    # If there is no ID column, then add one with a single value
+    if not df_has_id_column:
+        log.debug("Provided DataFrame (df) contains a single time series.")
```
LeonieFreisinger, line 52: I am wondering if this debug info is necessary or if we could leave it out. Just a detailed thought, however, feel free to decide based on your preferences.

neuralprophet/df_utils.py
```diff
@@ -36,31 +36,32 @@ def prep_or_copy_df(df):
...
+        df_copy["ID"] = "__df__"
+        return df_copy, df_has_id_column, True, ["__df__"]
+
+    # Create a list of unique ID values
+    unique_id_values: list[str] = df_copy["ID"].unique().tolist()
```
LeonieFreisinger, line 57: From my perspective, "unique" in de variable name is rather confusing. I rather see it as we are using the unique() method to get to the goal of having a list with all IDs present in the df.

neuralprophet/df_utils.py
```diff
@@ -36,31 +36,32 @@ def prep_or_copy_df(df):
...
-        raise ValueError("df is None")
-    else:
-        raise ValueError("Please, insert valid df type (pd.DataFrame)")
+    if not isinstance(df, pd.DataFrame):
+        raise ValueError("Provided DataFrame (df) must be of pd.DataFrame type.")
+
+    # Create a copy of the dataframe
+    df_copy = df.copy(deep=True)
+
+    df_has_id_column = "ID" in df_copy.columns
```
LeonieFreisinger, line 48: I very much like the variable naming  `df_has_id_column`

LeonieFreisinger (approved):
@noxan Great work in making it more readable. I left some comments, which definitely are very subject. Feel free to decide if you would like to address any of it.

neuralprophet/df_utils.py
noxan: just thought it might help people to identify which dataframe we mean 🤔

neuralprophet/df_utils.py
noxan: :)

neuralprophet/df_utils.py
noxan: We have a lot of logging information which I'm not sure how helpful it actually is, see #982 - as long as it is debug I would keep it for now (meaning it does not show by default anyhow)

neuralprophet/df_utils.py
noxan: Good point, I check yesterday that the copy operation is very lightweight on memory at least - would consider this change as part of #1167

PR #1185, opened 2023-03-01
[docs] Turn pull request prefixes to lower case

noxan:
- Change the pull request prefixes to lower case to align with current practice


judussoari (approved):
LGTM!

PR #1186, opened 2023-03-01
[devops] Downgrade python to 3.7 for pyright

noxan:
## :microscope: Background

- We support both python version, but we get new errors in python `3.10` which do not exist in `3.7` (`3.7` is most important for colab and other use cases, so focus on that)

## :crystal_ball: Key changes

- Change python version to `3.7` from `3.10` for pyright


judussoari (approved):
LGTM, thanks for fixing this!

PR #1187, opened 2023-03-02
[major] Allow user to set prediction frequency as a dict: {freq: forecast origin}

judussoari:
## :microscope: Background

Picks up from PR #1038 

Comment from reviewer:
Challenges: how is the start-time set? (as n_lags, and CV splits affect where the first sample starts, this is hard to be controlled implicityly)
Further, could there be a case where missing dates are not added, and thus gaps exists? Then, the approach to filter but every forecast_frequencyth value would fail as it would get shifted by the gaps.
An approach that does some sort of time modulo operation (if feasible), would likely address both issues.

Next PR: Forecast Gap! skip some forecasts (e.g. forecasting origin at 2pm, but skip the first 10 hours, and start forecasting at midnight next day)
Also another PR: Re-think the df_forecast in target-wise orientation. It may be easier to stick with raw (origin-wise) format, and more intuitive for the user

Allow user to specify:
- [x] when forecast are made (e.g. every Monday in case of daily data resolution), or Every day at 2pm (hourly data).. 

## :crystal_ball: Key changes

`prediction_frequency` is a dict with one key from `["daily","hourly","weekly","yearly"]` and an `int` value indicating at what hour of the day, minute of the hour, day of the week, and month of the year we would like to forecast. 
Example: `prediction_frequency = {"daily": 13}` indicated that we only train and forecast samples once per day, with the forecast origin fixed to 1pm. 

To DO:
- [x] Bei Trend and Season: Nanmean if we have overlapping/gappy trend and seas forecast when reshaping them to one trend/seas forecast column! 


neuralprophet/df_utils.py
```diff
@@ -1595,3 +1595,56 @@ def add_weekday_condition(df):
...
+            mask = ds.dt.month == target_time
+        elif key == "hourly":
+            target_time = target_time % 60
+            mask = ds.dt.minute == target_time
+        else:
+            raise ValueError(f"prediction_frequency {key} not supported")
+    yhat = np.full((len(ds),), np.nan)
+    yhat[mask] = forecast
+    yhat = np.concatenate(([np.NaN] * pad_before, yhat, [np.NaN] * pad_after))
+    return yhat
```
judussoari, line 1650: Added a new utils func that reshapes the predictions in m.predict() for the special case when the user opts to set a prediction frequency.

neuralprophet/forecaster.py
```diff
@@ -3242,7 +3244,24 @@ def _reshape_raw_predictions_to_forecst_df(self, df, predicted, components, pred
...
+                                freq = 30
+                            elif key == "hourly":
+                                freq = 60
+                            else:
+                                raise ValueError(f"Unknown prediction frequency {key}")
+                            # Check if the next forecast for the component overlaps with the previous one
+                            if self.n_forecasts > freq:
+                                forecast_rest = components[comp][1:, -(self.n_forecasts - freq) :, j]
+                            else:
+                                forecast_rest = components[comp][1:, :, j]
```
judussoari, line 3264: For Trend, Seasonality and so on: Flatten the forecasts for those components according to the prediction frequency

neuralprophet/time_dataset.py
```diff
@@ -176,7 +180,27 @@ def filter_samples_after_init(
...
+            elif key == "hourly":
+                mask = timestamps.minute == value + 1
+            else:
+                raise ValueError(f"Invalid prediction frequency: {key}")
+        self.samples = [self.samples[i] for i in range(len(self.samples)) if mask[i]]
+
+        # Exact timestamps are not needed anymore
+        self.inputs.pop("timestamps")
+        for sample in self.samples:
+            sample.pop("timestamps")
```
judussoari, line 203: Filtering the samples according to the prediction frequency. See description of the PR to see how the dictionary prediction_frequency works. 
To achieve this, the real timestamps are passed to the samples, but then dropped after filtering is done as we only need them to filter the samples and for nothing else. This keeps the samples lean

neuralprophet/df_utils.py
```diff
@@ -1602,3 +1602,56 @@ def add_weekday_condition(df: pd.DataFrame):
...
+            identical to NeuralProphet
+        forecast_lag : int
+            identical to NeuralProphet
+
+    Returns
+    -------
+        np.array
+            reshaped yhat array
+    """
+    for key, value in prediction_frequency.items():
```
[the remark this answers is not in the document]
judussoari: Done

neuralprophet/forecaster.py
```diff
@@ -3288,12 +3296,30 @@ def _reshape_raw_predictions_to_forecst_df(self, df, predicted, components, pred
...
+                            for date in dates_comp:
+                                d = pd.date_range(date, periods=self.n_forecasts + 1, freq=self.data_freq)
+                                ser = pd.concat((ser, pd.Series(d).iloc[1:]))
+                            df_comp = pd.DataFrame({"ds": ser, "yhat": components[comp].flatten()}).drop_duplicates(
+                                subset="ds"
+                            )
+                            df_comp, _ = df_utils.add_missing_dates_nan(df_comp, freq=self.data_freq)
+                            yhat = pd.merge(df_forecast.filter(["ds", "ID"]), df_comp, on="ds", how="left")[
+                                "yhat"
+                            ].values
```
judussoari, line 3322: For any case (n_forecasts <,>,= prediction_frequency) the non-lagged components are now returned in a clean way (dropping duplicate trend predictions, or adding missing NaNs in case of n_forecasts > prediction_freq, and n_forecasts < prediction_freq, respectively)

neuralprophet/forecaster.py
```diff
@@ -3288,12 +3296,30 @@ def _reshape_raw_predictions_to_forecst_df(self, df, predicted, components, pred
...
                     yhat = np.concatenate(([np.NaN] * self.max_lags, forecast_0, forecast_rest))
                     if prediction_frequency is not None:
-                        forecast_rest = components[comp][1:, :, j]
-                        yhat = np.concatenate(([np.NaN] * self.max_lags, forecast_0, forecast_rest.flatten()))
-                        if len(yhat) < len(df_forecast):
-                            yhat = np.concatenate((yhat, [np.NaN] * (len(df_forecast) - len(yhat))))
-                        else:
-                            yhat = yhat[: len(df_forecast)]
+                        for key, value in prediction_frequency.items():
+                            if key == "daily":
```
ourownstory, line 3300: To make the unit to be provided clear, I suggest to add it to the key, e.g. "daily-hour"

neuralprophet/forecaster.py
```diff
@@ -296,12 +296,12 @@ class NeuralProphet:
...
             Dictionary of additional trainer configuration parameters.
-        prediction_frequency: int
+        prediction_frequency: dict
             periodic interval in which forecasts should be made.
-
-            Note
-            ----
-            E.g. if prediction_frequency=7, forecasts are only made on every 7th step (once in a week in case of daily resolution).
+            Key: str
+                periodicity of the predictions to be made, e.g. 'daily'.
```
ourownstory, line 302: Please add all available options (as docstring `Options`)

neuralprophet/forecaster.py
```diff
@@ -296,12 +296,12 @@ class NeuralProphet:
             Provide `None` to deactivate the use of accelerators.
         trainer_config: dict
             Dictionary of additional trainer configuration parameters.
-        prediction_frequency: int
+        prediction_frequency: dict
             periodic interval in which forecasts should be made.
```
ourownstory, line 300: Specify how multiple items in a dict are interpreted

ourownstory (approved):
Looking great!
I suggest minor renamings and addition of test
- unit added to key
-  docstring: Specify how multiple items in a dict are interpreted
- docstring: Please add all available options (as docstring Options)
- tests: include a combination of keys

PR #1188, opened 2023-03-02
[minor] Support data without datestamps

### LINKED ISSUE #18: Non-date data: Support equidistant data without datestamps
Currently, the best approach is to simply map it to some random series of datestamps.
A helper function to do this might be useful, as dealing with datetimes can be messy.

However, proper support for such data should be implemented eventually. It is not complex, but somewhat complicated as there are many touch points with datetime stamps.
This will entail:
* data checks and missing data imputation
* trend and seasonality computation
* plotting support

SimonWittner:
## :microscope: Background

- Processing equidistant-data without datestamps
- Closes #18 

## :crystal_ball: Key changes

- Created a helper function to add equidistant datestamps 
- Function can be manually called and adjusted to the required dates and frequency 
- Function is automatically called if df consists only of column "y"


karl-richter:
Hi @SimonWittner, I know you are just getting started with this pr - but maybe have a look at [this](https://github.com/ourownstory/neural_prophet/blob/main/CONTRIBUTING.md). This will make it easier to have meaningful commit messages, pr titles and so on :)

Also, the model/metrics pytest does currently not work from forks yet. You can it if that pytest fails for now.

neuralprophet/df_utils.py
```diff
@@ -37,11 +36,15 @@ def prep_or_copy_df(df: pd.DataFrame) -> tuple[pd.DataFrame, bool, bool, list[st
         bool
             wheter it is a single time series
         list
-            list of IDs
+            list of IDsgithubn
```
judussoari, line 39: ?

ourownstory (changes requested):
Thank you @SimonWittner for your contribution!

While I was thinking of adding true support for data without datestamps, which would be entail large codebase changes, your approach is a rather simple alternative that may get the job done. I think this is a good addition, so I'd like to proceed to get your contribution merge-ready.
To do so, I suggest to still do the following:
- rename mentions of 'random' to 'dummy' or 'placeholder'
- use seconds as the frequency (to avoid issues with very long time-series)
- make sure to disable auto-seasonality when your code is triggered. (as the daily/weekly/yearly seasonalities will have no meaning)
- May there be a better location (less nested and less frequently called) than `prep_or_copy_df` for your code to be called?
- Add to the user warning that the frequency was set to seconds
- soft suggestion: have the function only return a series of datestamps, (no passing of df), and use a more efficient implementation of adding the column to the df than concat (outside the function).

Please let me know if you disagree with my suggestions or have further ideas

neuralprophet/df_utils.py
leoniewgnr: sollte da nicht `IDs` stehen?

leoniewgnr (changes requested):

leoniewgnr (changes requested):
Looks good so far but I think we should also support m.make_future_dataframe

leoniewgnr (approved):
LGTM

neuralprophet/df_utils.py
```diff
@@ -513,6 +511,42 @@ def check_single_dataframe(df, check_y, covariates, regressors, events, seasonal
     return df
 
 
+def create_dummy_datestamps(
```
ourownstory, line 514: Looks good overall!

Let us separate the disabling (editing of model config) from the  timestamps creation.

So, for this function, let us not pass the model, but only the precise args that are needed.

ourownstory (changes requested):
Good changes!

I noticed that the model is passed to the datestamp creation function. 
Let us separate the disabling (editing of model config) from the timestamps creation.
So, for datestamp creation function, let us not pass the model, but only the precise args that are needed.

Further, the editing of model parameters (in a separate function), should not happen in `make_future_dataframe`. If seasonalities are enabled there and no datestamps present, it should create an error.

SimonWittner:
Handling of data without datestamp fixed with https://github.com/ourownstory/neural_prophet/pull/1432

PR #1193, opened 2023-03-04
[typing] Add typing for user facing functions

### LINKED ISSUE #1134: Add typing to all user-facing functions
- [ ] scope 1 : Forecaster.py
- [ ] scope 2: all user-facing

noxan:
## :microscope: Background

Close #1134

### DIFF
--- a/neuralprophet/forecaster.py
+++ b/neuralprophet/forecaster.py
@@ -584,7 +584,14 @@ class NeuralProphet:
         self.config_regressors[name] = configure.Regressor(reg_lambda=regularization, normalize=normalize, mode=mode)
         return self
 
-    def add_events(self, events, lower_window=0, upper_window=0, regularization=None, mode="additive"):
+    def add_events(
+        self,
+        events: Union[str, list],
+        lower_window: int = 0,
+        upper_window: int = 0,
+        regularization: Optional[float] = None,
+        mode: str = "additive",
+    ):
         """
         Add user specified events and their corresponding lower, upper windows and the
         regularization parameters into the NeuralProphet object
--- a/neuralprophet/forecaster.py
+++ b/neuralprophet/forecaster.py
@@ -10,6 +10,7 @@ import pandas as pd
 import pytorch_lightning as pl
 import torch
 from matplotlib import pyplot
+from matplotlib.axes import Axes
 from torch.utils.data import DataLoader
 
 from neuralprophet import configure, df_utils, np_types, time_dataset, time_net, utils, utils_metrics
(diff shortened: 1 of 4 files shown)


neuralprophet/forecaster.py
```diff
@@ -1295,7 +1322,14 @@ def crossvalidation_split_df(
                 del folds[i][1]["ID"]
         return folds
 
-    def double_crossvalidation_split_df(self, df, freq="auto", k=5, valid_pct=0.10, test_pct=0.10):
+    def double_crossvalidation_split_df(
+        self,
+        df: pd.DataFrame,
+        freq: str = "auto",
+        k: int = 5,
+        valid_pct: float = 0.10,
```
LeonieFreisinger, line 1330: from a consistency perspective, we can remove the last 0.

neuralprophet/forecaster.py
```diff
@@ -1372,7 +1406,14 @@ def create_df_with_events(self, df, events_df):
...
         return df
 
-    def make_future_dataframe(self, df, events_df=None, regressors_df=None, periods=None, n_historic_predictions=False):
+    def make_future_dataframe(
+        self,
+        df: pd.DataFrame,
+        events_df: Optional[pd.DataFrame] = None,
+        regressors_df: Optional[pd.DataFrame] = None,
+        periods: Optional[int] = None,
+        n_historic_predictions: Union[bool, int] = False,
```
LeonieFreisinger, line 1415: We should rethink in the future if we allow int variables like `n_historic_predictions` to have bool default. Is there a best practise how to prevent that?

neuralprophet/forecaster.py
```diff
@@ -1444,7 +1485,12 @@ def make_future_dataframe(self, df, events_df=None, regressors_df=None, periods=
         )
         return df_future
 
-    def handle_negative_values(self, df, handle="remove", columns=None):
+    def handle_negative_values(
+        self,
+        df: pd.DataFrame,
+        handle: Union[str, int, float, None] = "remove",
```
LeonieFreisinger, line 1491: `handle`  could be assigned 4 different types. Would be interested if there are best practices to simplify that piece of code.

LeonieFreisinger (approved):
@noxan Thanks for adding typing to the relevant functions! Great effort.

neuralprophet/forecaster.py
noxan: not related this PR, but happy to do so :)

neuralprophet/forecaster.py
noxan: We should make our default parameter selection more consistent. Currently we have `None`, `False`, `auto`, etc - but not related to this PR.

neuralprophet/forecaster.py
noxan: We could wrap them in a custom type, not sure if that improves readability though... would leave it as this. Question is rather why do we accept that many different types in the first place.

Issue #1201, opened 2023-03-07
Review and remove unused dependencies


noxan:
It appears some dependencies of our project might not be used at all. Go over them and check if they are used, if not remove them.

Good preparation for #911

noxan:
## Legend

- [!] For dependency not being used
- [ ] For dependency to be checked
- [x] For dependency which is being used (import)

## Base requirements

- [ ] captum>=0.5.0
- [ ] convertdate>=2.1.2
- [ ] dataclasses>=0.6;python_version<'3.7'
- [ ] holidays>=0.20
- [ ] ipywidgets>=7.5.1
- [ ] LunarCalendar>=0.0.9
- [ ] matplotlib>=2.0.0
- [ ] numpy>=1.15.4,<1.24.0
- [ ] pandas>=1.0.4
- [ ] plotly>=5.12.0
- [ ] plotly_resampler>=0.8.3
- [ ] python-dateutil>=2.8.0
- [ ] pytorch-lightning>=1.7
- [ ] tensorboard
- [ ] torch>=1.8.0
- [ ] torchmetrics>=0.9.3
- [ ] typing_extensions>=4.4.0;python_version<'3.8'
- [ ] widgetsnbextension>=4.0.5

Issue #911, opened 2022-10-27
Project dependencies may have API risk issues


PyDeps:
Hi, In **neural_prophet**, inappropriate dependency versioning constraints can cause risks.

Below are the dependencies and version constraints that the project is using

```
numpy>=1.15.4
pandas>=1.0.4
matplotlib>=2.0.0
torch>=1.8.0
LunarCalendar>=0.0.9
convertdate>=2.1.2
holidays>=0.11.3.1
python-dateutil>=2.8.0
tqdm>=4.50.2
torch-lr-finder>=0.2.1
ipywidgets>=7.5.1
plotly>=4.14.3
dataclasses>=0.6;python_version<'3.7'
```

The version constraint **==** will introduce the risk of dependency conflicts because the scope of dependencies is too strict.
The version constraint **No Upper Bound** and **\*** will introduce the risk of the missing API Error because the latest version of the dependencies may remove some APIs.

After further analysis, in this project,
The version constraint of dependency **matplotlib** can be changed to *>=1.4.0,<=3.0.3*.
The version constraint of dependency **holidays** can be changed to *>=0.2,<=0.14.2*.
The version constraint of dependency **python-dateutil** can be changed to *>=2.1,<=2.8.2*.
The version constraint of dependency **tqdm** can be changed to *>=4.36.0,<=4.64.0*.

The above modification suggestions can reduce the dependency conflicts as much as possible, 
and introduce the latest version as much as possible without calling Error in the projects.

The invocation of the current project includes all the following methods.

<details><summary>The calling methods from the matplotlib</summary>
 <pre>matplotlib.dates.MonthLocator
matplotlib.dates.num2date
matplotlib.dates.AutoDateLocator
matplotlib.dates.AutoDateFormatter
matplotlib.ticker.FuncFormatter
</pre>
</details>
<details><summary>The calling methods from the holidays</summary>
 <pre>holidays.HolidayBase.__init__
</pre>
</details>
<details><summary>The calling methods from the python-dateutil</summary>
 <pre>dateutil.relativedelta.relativedelta
dateutil.easter.easter
</pre>
</details>
<details><summary>The calling methods from the tqdm</summary>
 <pre>tqdm.tqdm
</pre>
</details>
<details><summary>The calling methods from the all methods</summary>
 <pre>plot_custom_season
events_df.copy.reset_index
i.self.ar_net
neuralprophet.plot_forecast_plotly.plot_components
df.reset_index.copy
[...]

Issue #1129, opened 2023-01-24
[Inclusion of Neural Prophet in Darts] relaxing dependencies


hrzn:
Hi team, @ourownstory, 
We are considering introducing Neural Prophet in Darts (see: https://github.com/unit8co/darts/pull/1436). Unfortunately the current requirements are a bit too strict for us. Both pytorch lightning and torchmetrics versions are currently pinned. I made a quick test and it seems your unit tests are running fine with Pytorch lightning 1.9.0. Would it be possible to consider relaxing the dependencies (at least with `>=`, and if possible with older lower bounds)? That could maybe also benefit some of your users.

Cheers,
Julien

karl-richter:
Hi Julien, thanks so much for considering the integration of NP into Darts.
I'm looking into the dependencies rn, we can definitely loosen them by a lot! At some point we ran into troubles with a certain version, but that should be fixed now. You can track the progress in #1145.

Best, Karl

karl-richter:
@hrzn, we adjusted the dependencies after some evaluations of the newer package versions. Hope that helps. Let me know if you notice any further roadblocks.

hrzn:
Very nice, looks great, thanks @karl-richter !

PR #1202, opened 2023-03-07
[minor] Dependency management with poetry

noxan:
## :microscope: Background

- [Poetry](https://python-poetry.org/) is a modern python dependency manager unifying all requirements and distribution related configuration

## :crystal_ball: Key changes

- Review all of our dependencies and remove unused ones (#1201)
- Lock all package version numbers to their latest feature release which is compatible with our minimum python version of `3.7.1` (#911)

## ❓ Discussion points

1. Any reasons not to remove the `requirements/*.txt` setup?

## 📓 Open tasks

1. Migrate the github actions to the new poetry setup (away from pip)

## ✨ Why poetry?

1. It ensures package versions are compatible with the python version
2. Ensures the right package dependency version and their dependencies are installed with its lock-file
3. Automatically manages the virtual environment setup
4. Integrates the publish workflow (pypi)
5. Supports different dependency groups (e.g. `tests`, `docs`) and unifies them all into a single configuration file
6. Easily keep track of all outdated packages (with `poetry list --outdated`) and update them accordingly


ourownstory:
@noxan I concur that this would be the right thing to do if our package would not be used within other packages or in environments with other packages. 
However, tightening the requirements more than critically necessary will cause issues for libraries such as Darts using our package, as you can see from https://github.com/ourownstory/neural_prophet/issues/1129
Also, users who want to use multiple different models in the same environment may run into issues.

Does poetry allow us to relax, rather than tighten the dependency bounds?

A small point is the familiarity most developers/users have with pip vs poetry.

noxan:
> However, tightening the requirements more than critically necessary will cause issues for libraries such as Darts using our package, as you can see from https://github.com/ourownstory/neural_prophet/issues/1129
Also I agree with your argument of not defining them to tightly as it creates issues. But having unbound dependency versions creates other kinds of issues, e.g. our package breaks if another dependency introduces a breaking change. And in my experience it is better to be explicit about versions, it is easy to report a too tightly bound dependency and relax it. In contrary you can spend hours on figuring out which versions are incompatible, and why which package breaks. Especially if they are not explicitly defined, everything works fine, you recreate your development environment or a new developer joins the team and nothing works anymore - because the dependency versions luckily overlapped during your first project setup, just now things changed and nobody has a clue what...

> Does poetry allow us to relax, rather than tighten the dependency bounds?
How tightly we define our dependencies is independent of poetry or pip. They both allow the same options.

>  Also, users who want to use multiple different models in the same environment may run into issues.
Into what kind of issues? That package versions are not compatible? At least it would be explicit then, good look trying to figure out which exact version of a dependency (3rd party) of your dependencies are incompatible to another.

> A small point is the familiarity most developers/users have with pip vs poetry.
Agree - poetry also makes a lot of things a lot easier

noxan:
@ourownstory 

> However, tightening the requirements more than critically necessary will cause issues for libraries such as Darts using our package, as you can see from https://github.com/ourownstory/neural_prophet/issues/1129

Also I agree with your argument of not defining them to tightly as it creates issues. But having unbound dependency versions creates other kinds of issues, e.g. our package breaks if another dependency introduces a breaking change. And in my experience it is better to be explicit about versions, it is easy to report a too tightly bound dependency and relax it. In contrary you can spend hours on figuring out which versions are incompatible, and why which package breaks. Especially if they are not explicitly defined, everything works fine, you recreate your development environment or a new developer joins the team and nothing works anymore - because the dependency versions luckily overlapped during your first project setup, just now things changed and nobody has a clue what...


> Does poetry allow us to relax, rather than tighten the dependency bounds?

How tightly we define our dependencies is independent of poetry or pip. They both allow the same options.


>  Also, users who want to use multiple different models in the same environment may run into issues.

Into what kind of issues? That package versions are not compatible? At least it would be explicit then, good look trying to figure out which exact version of a dependency (3rd party) of your dependencies are incompatible to another.


> A small point is the familiarity most developers/users have with pip vs poetry.

Agree - yet poetry also makes a lot of things a lot easier

noxan:
Another thing which I became aware of: we use a lot of dependencies which are actually not supporting our python version anymore, e.g. pandas dropped python `3.7` support with version `1.4`. Our requirements allows anything `pandas>=1.0.4` - apparently it still works but how knows for how long they break with python `3.7` - it is not even officially supported anymore.

LeonieFreisinger (approved):
LGTM @noxan Please have a look at the failing test again

PR #1203, opened 2023-03-07
[minor] Adjust seasonality reg to modularized code

judussoari:
## :microscope: Background

Bug because `self.season_params` does not exist anymore with the newly modularized seasonality. 

## :crystal_ball: Key changes

Change to `self.seasonality.season_params`


karl-richter (approved):
Thanks for fixing this!

PR #1215, opened 2023-03-14
[feature] Support for asymmetrical interval in conformal quantile regression

leoniewgnr:
## :microscope: Background
Until now, it was only possible to add only symmetrical intervals for conformal quantile regression, e.g. [0.05, 0.95]

## :crystal_ball: Key changes
In `uncertainty.py`, I changed `q_hat` to `q_hat_lo` or `q_hat_h`i, so that there is now a different prediction interval for upper and lower uncertainty if an asymmetrical interval was given (e.g. [0.1, 0.95])
Also fixed the plot function in `uncertainty.py` so that it is able to show asymmetrical intervals. 

Already before my changes some plot functions for conformal predictions are not working. I marked them with a comment 
`# plot not working yet`


leoniewgnr:
It is now possible to control the error coverage on each tail respectively. Just define α=(0.03, 0.07) as a tuple of floats for left and right tail instead of single float (α = 0.1). Setting a single float as α is still possible, then the coverage error is spread arbitrarily over left and right tails. Implement according to the[ CQR Whitepaper](https://arxiv.org/abs/1905.03222).
Asymmetrical CQR

neuralprophet/uncertainty.py
```diff
@@ -35,11 +35,26 @@ class Conformal:
...
+        if isinstance(self.alpha, float):
+            self.symmetrical = True
+            self.q_hats = pd.DataFrame(columns=["q_hat_sym"])
+        elif self.method == "naive":
+            raise ValueError(
+                "Asymmetrical coverage errors are not available for the naive method. "
+                "Please use one alpha or method='cqr'."
+            )
+        else:
+            self.symmetrical = False
```
Kevin-Chen0, line 53: For CQR, should we give the user the option to pick whether `symmetrical` or not? I think it should be, therefore `symmetrical` needs to be passed in as a param. As `alpha` should be independent of symmetry. One float `alpha` can be either symmetric (combine noncon scores) or asymmetric (hi and lo scores) while tuple `alpha` can also be symmetric or asymmetric. `alpha` simply means where to extract the `q_hat` value(s) from the noncon scores.

Therefore, I think that `symmetrical` should be param in with the default set to True. If `method` is 'naive' yet symmetrical is False, then it throws an error because naive can only be symmetrical atm.

neuralprophet/uncertainty.py
```diff
@@ -56,38 +71,59 @@ def predict(self, df: pd.DataFrame, df_cal: pd.DataFrame) -> pd.DataFrame:
...
             noncon_scores = self._get_nonconformity_scores(df_cal, step_number)
             q_hat = self._get_q_hat(df_cal, noncon_scores)
-            q_hat_col = f"qhat{step_number}"
             y_hat_col = f"yhat{step_number}"
-            df[q_hat_col] = q_hat
-            if self.method == "naive":
-                df[f"{y_hat_col} - {q_hat_col}"] = df[y_hat_col] - q_hat
-                df[f"{y_hat_col} + {q_hat_col}"] = df[y_hat_col] + q_hat
-            elif self.method == "cqr":
+            if self.method == "naive" and self.symmetrical:
```
Kevin-Chen0, line 80: So it is possible to be CQR is symmetrical right?

Kevin-Chen0 (changes requested):

neuralprophet/uncertainty.py
leoniewgnr: I think we should not pass as `symmetrical` as an option to the user, but we should let the user choose to use `symmetrical` or not by either setting `alpha` to a float (symmetrical) or tuple (asymmetrical).
Doing it this way, it is not possible to use asymmetric (hi and lo scores) with a symmetric alpha, but this is intended because of two reasons:
1. In all the literature, I found nobody, who would do that. The Whitepaper use asymmetric (hi and lo scores) always with asymmetric alpha.
2. It is already a very specific use case, if anyone requests even more specifically, I'm happy to change it. But for now, this is good for most of the use cases

Kevin-Chen0: Let's say, in a hypothetical example, the lower QR contains sorted nonconformity score of [1, 2, 3, 4, 5, 6, 7, 8, 9, 10] and the upper QR contains that of [2, 4, 6, 8, 10, 12, 14, 16, 18, 20]. Thus, the combination is [1, 2, 2, 3, 4, 4, 5, 6, 6, 7, 8, 8, 9, 10, 10, 12, 14, 16, 18, 20]. If alpha = 0.1, and you set it as symmetric, then symm q-hat will be **16**. If asymmetric but alpha for both is 0.1, then lo q_hat is **9** and hi q_hat is **18**. So it is very possible to have only one alpha but two different q-hats for lower and upper. Conversely, if alpha = (0.1, 0.2), then lower is  q_hat is **9** and hi q_hat is **16**.

Kevin-Chen0: That said, in your code, the user don't have to pass `symmetric` keyword as long as the tuple automatically signifies asymmetry (which I presume it does now) _and_  it accepts two of the same alpha values. For example, **(0.1, 0.1)**.

leoniewgnr: **First comment**
In [_get_nonconformity_scores](https://github.com/ourownstory/neural_prophet/blob/main/neuralprophet/uncertainty.py#:~:text=def-,_get_nonconformity_scores,-(self%2C), there are not lower and upper QR for symmetrical CQR. First it takes the max (distance to lower quantile, distance to upper quantile) and only afterward sorts them. 
But yes, for the asymmetric, this true.

**Second comment**
Yes, thats correct. So if you want to reach that behavior, you could just set alpha = (0.1, 0.1)

But then its fine, isnt it? Or what would you like to change?

Kevin-Chen0: Sounds good. Nothing needs to be changed for this part.

Kevin-Chen0 (approved):
LGTM

PR #1216, opened 2023-03-14
[major] Nested Additive Model

karl-richter:
## Background

- The current purely additive model is theoretically not correct when autoregressive components are used.

## Key changes

- Introduced a nested additive procedure that first learns all non-stationary components, and the AR-Net on the remainder.
- For this implementation, the non-stationary components include trend, seasonality, events, and future regressors.

## Review Checklist


judussoari (approved):
Awesome work @LeonieFreisinger @karl-richter , this is gonna be super helpful! 
Changes look great. For a subsequent PR we might consider:

- Come up with a way to decrease the number of function calls for the nested additive model (runtime)
- Handle quantiles

Issue #1235, opened 2023-03-27
Add plotly-static plotting backend


noxan:
**Is your feature request related to a problem? Please describe.**

- The plots in the tutorial Jupyter notebooks cannot be saved and restored due the interactive nature of plotly plots.

**Describe the solution you'd like**

- Add a `plotly-static` option for our plotting backend which renders static (non-interactive) plotly plots.

**Alternatives considered**

- It would be possible to set the plotting backend to `plotly` (without the resampler), capture the returned figure and override the renderer, e.g. `fig.show(renderer="svg")`. But this is rather unpractical if users want to opt for the interactive version when they work through the tutorials and have to adjust this each and every time.

hxyue1:
I can work on this

PR #1259, opened 2023-04-13
# Plolty plot didn't show on colab

vijaybirju:
Plotly plot didn't show on Colab

## :microscope: Background

- This change will let show plolly plots on colab notebook
## :crystal_ball: Key changes

- Explain the main changes introduced by this pull request for the reviewer.

I have added a code shell with following code 
from plotly.offline import init_notebook_mode, iplot
init_notebook_mode(connected=True)


leoniewgnr (changes requested):
Hi @vijaybirju, thanks so much for taking this over. Just a question: This change is only for google colab relevant, isn't it? So should we put the two additional lines in the first if-clause, where we check if google.colab is used?

noxan:
I can confirm the same issue with interactive plots in jupyter notebooks locally in vs code - but it's an issue to all notebooks not only this one in particular - #1235

vijaybirju:
it 

> Hi @vijaybirju, thanks so much for taking this over. Just a question: This change is only for google colab relevant, isn't it? So should we put the two additional lines in the first if-clause, where we check if google.colab is used?

it is for Colab and Kaggle. Should I check for both or just for colab

vijaybirju:
@leoniewgnr can you review it now?

vijaybirju:
> I can confirm the same (or at least a similar) issue with interactive plots in jupyter notebooks locally in vs code - but it's an issue to all notebooks not only this one in particular - #1235

yes it happened in Colab and Kaggle environment

leoniewgnr:
@vijaybirju sorry for not coming back to you, please feel free to ping me on slack, I'll respond faster. 
Thanks so much for taking this one, I changed the location of the tutorials (tidying up the docs folder structure). So the notebook doesn't exist anymore at the current location, so I just put your changes in the notebook of the new location.
See https://github.com/ourownstory/neural_prophet/pull/1309

Issue #1261, opened 2023-04-13
[Refactor] Streamline uncertainty naming of quantile


leoniewgnr:
Right now, every new uncertainty estimation is added as a new column in the forecast dataframe. See [here](https://github.com/ourownstory/neural_prophet/blob/658347318cd7106268e4128b3b6ed260c9fbf391/neuralprophet/uncertainty.py#L75). It would be much cleaner, if old uncertainty intervals would be overwritten because they are not needed anymore if a better one is computed.

PR #1262, opened 2023-04-13
[refactor] Uncertainty streamline naming

leoniewgnr:
## :microscope: Background
The forecast data frames gets a new column for every new uncertainty prediction method which leads to a very large data frame. 
Fixes issue #1261 

## :crystal_ball: Key changes
The initial uncertainty intervals are overwritten with the new ones when a new uncertainty prediction method is used. 


neuralprophet/uncertainty.py
```diff
@@ -77,39 +78,21 @@ def predict(self, df: pd.DataFrame, df_cal: pd.DataFrame) -> pd.DataFrame:
...
-                df[q_hat_col] = q_hat_sym
-                quantile_lo = str(min(self.quantiles) * 100)
-                quantile_hi = str(max(self.quantiles) * 100)
-                quantile_lo_col = f"{y_hat_col} {quantile_lo}%"
-                quantile_hi_col = f"{y_hat_col} {quantile_hi}%"
-                df[f"{quantile_lo_col} - {q_hat_col}"] = df[quantile_lo_col] - q_hat_sym
-                df[f"{quantile_lo_col} + {q_hat_col}"] = df[quantile_lo_col] + q_hat_sym
-                df[f"{quantile_hi_col} - {q_hat_col}"] = df[quantile_hi_col] - q_hat_sym
-                df[f"{quantile_hi_col} + {q_hat_col}"] = df[quantile_hi_col] + q_hat_sym
+                df[y_hat_lo_col] = df[y_hat_col] - q_hat_sym
```
Kevin-Chen0, line 89: Needs to be:

`df[y_hat_lo_col] = df[y_hat_lo_col] - q_hat_sym`

As you are replacing the existing QR's lower quantile with that of CP. Also apply with this to upper quantile and assymetrical CQR.

Kevin-Chen0 (changes requested):

neuralprophet/uncertainty.py
```diff
@@ -77,39 +78,21 @@ def predict(self, df: pd.DataFrame, df_cal: pd.DataFrame) -> pd.DataFrame:
...
-                q_hat_col = f"qhat{step_number}"
-                df[q_hat_col] = q_hat_sym
-                df[f"{y_hat_col} - {q_hat_col}"] = df[y_hat_col] - q_hat_sym
-                df[f"{y_hat_col} + {q_hat_col}"] = df[y_hat_col] + q_hat_sym
+                df[y_hat_lo_col] = df[y_hat_col] - q_hat_sym
+                df[y_hat_hi_col] = df[y_hat_col] + q_hat_sym
             elif self.method == "cqr" and self.symmetrical:
                 q_hat_sym = q_hat["q_hat_sym"]
-                q_hat_col = f"qhat{step_number}"
-                df[q_hat_col] = q_hat_sym
```
Kevin-Chen0, line 89: Idea is that for the advanced users, they need to input the `q_hat` into the forecast df themselves using the noncon scores right?

neuralprophet/forecaster.py
```diff
@@ -3359,7 +3359,7 @@ def conformal_predict(
             quantiles=self.config_train.quantiles,
```
Kevin-Chen0, line 3359: I'll be honest, the current structure here makes it basically unusable for the advanced user. That is because the advanced user has to basically call the Conformal class, get the noncon scores, use the noncon scores to get the q_hat, then put the q_hat into df. Those are all supposed to be behind-the-scenes in the `Conformal` method, regardless of the user.

I'd strongly prefer rather than all of that, you instead add one more param in the `conformal_predict()`, something like `show_conformal` etc. `False` is for the basic user (and set to default) and `True` is for the advanced user. If set to True, then the output of `df_forecast` is you currently have: only the QR columns. If set to False, then the output of `df_forecast` is what I currently have prior to this PR: columns with both QR and CP/CQR PIs.

Basically, there will be two workflows, one for basic and one for advanced. But that's ok as these workflows should be encapsulated within the `Conformal` class and `conformal_prediction()` will only pass whatever `Conformal` outputs.

With this, a couple more things:
1. `Conformal.predict()` should only output `df_forecast` and no need to output `noncon` scores, as the current setup.
2. `Conformal.plot()` should only be shown for advanced users. Something like:

```
if show_conformal and plotting_backend:
    c.plot(plotting_backend)
```

Kevin-Chen0 (changes requested):

neuralprophet/uncertainty.py
leoniewgnr: good catch, adressed in most recent commit

neuralprophet/forecaster.py
leoniewgnr: Oke I understand, makes sense. I just want to avoid having the complete code duplicated, as this would mean doubled potential for bugs and double maintenance effort. Do you have an idea on how to avoid duplicated code?

leoniewgnr:
**Basic user:**
```
method = "naive"  # or "cqr" for a more sophisticated method, see uncertainty tutorial
conformal_forecast = m.conformal_predict(train_df, cal_df, alpha=0.1, method=method)
fig = m.highlight_nth_step_ahead_of_each_forecast(1).plot(conformal_forecast)
```
--> returns only y, yhat and the updated uncertainty intervals
--> plot shows only the forecast and one uncertainty interval around it

**Advanced user:**
```
method = "naive"  # or "cqr" for a more sophisticated method, see uncertainty tutorial
conformal_forecast = m.conformal_predict(train_df, cal_df, alpha=0.1, method=method, show_qr=True)
fig = m.conformal_plot(conformal_forecast, n_highlight =1)
```
--> returns updated uncertainty intervals (conformal prediction: naive or cqr) and previous quantile regression intervals
--> plot shows previous qr intervals in light blue and additional lines for conformal prediction

neuralprophet/forecaster.py
```diff
@@ -2848,17 +2852,73 @@ def conformal_predict(
...
-        df_forecast = c.predict(df=df_test, df_cal=df_cal)
+
+        if not show_qr:
+            # call Conformal's predict to output test df with conformal prediction intervals
+            df_forecast = c.predict(df=df_test, df_cal=df_cal)
+            # plot one-sided prediction interval width with q
+        else:
+            df_forecast = c.predict(df=df_test, df_cal=df_cal)
+
+            # get quantile regression intervals
```
Kevin-Chen0, line 2872: All of these modifications for advanced users (whatever they may be) should be done within the `c.predict()`. `conformal_predict()` should be as lightweight as the forecaster.py file is already long enough. You can create an additional helper function in Conformal class for advanced users and have `c.predict()` call if it helps not make it too overloaded.

neuralprophet/forecaster.py
```diff
@@ -2848,17 +2852,73 @@ def conformal_predict(
...
+            df_quantiles = df_qr.columns[df_qr.columns.str.contains("%")]
+            df_add = df_qr[df_quantiles]
+            df_add.columns = [f"qr_{col}" for col in df_add.columns]
+            df_forecast = pd.concat([df_forecast, df_add], axis=1, ignore_index=False)
```
Kevin-Chen0, line 2876: For naive CP without `show_qr` (basic users), the columns are `['ds', 'y', 'yhat1', 'yhat1 5.0%', 'yhat1 95.0%', 'trend',
       'season_weekly', 'season_daily']`. Where `yhat1 5.0%`, and `yhat1 95.0%` are the CP PI that replaced the QR ones (if existed). That's fine, as basic users only care about the PI regardless of the technique.

However, for naive CP with `show_qr` (advanced users), the columns are now `['ds', 'y', 'yhat1', 'yhat1 5.0%', 'yhat1 95.0%', 'trend',
       'season_weekly', 'season_daily', 'qr_yhat1 5.0%', 'qr_yhat1 95.0%']`. That is very confusing for me because it took a while for me to register that `yhat1 5.0%` and `yhat1 95.0%` refers to the CP PI. Therefore, I suggest that for advanced users, you use my existing naming convention. 

- `yhat1 <quantile>` for QR PI
- `yhat1 +- qhat1` for naive CP PI
- `yhat1 <quantile> +- qhat1` for CQR PI

I have this naming convention for a reason; it easily distinguishes the different PIs from each other. And it especially distinguishes naive CP and CQR as their qhats are used on the mean and the QR quantiles respectively. Therefore, their qhats are not used in the same way hence the difference naming columns.

Because of this, I also suggest the change the naming of param `show_qr` to something else. Like show all PIs and show complete uncertainty etc.

neuralprophet/forecaster.py
```diff
@@ -2804,6 +2805,7 @@ def conformal_predict(
         alpha: Union[float, Tuple[float, float]],
         method: str = "naive",
         plotting_backend: Optional[str] = None,
+        show_qr: Optional[bool] = False,
```
Kevin-Chen0, line 2808: Recommend changing the naming of param `show_qr` to something else. Like show all PIs or show complete uncertainty or for advanced users etc. See comment below.

Kevin-Chen0 (dismissed):

neuralprophet/forecaster.py
leoniewgnr: I implemented all demanded changes, but for this one, I want to outline the consequences this one would have.
If the naming goes back to `yhat <quantile>` (qr), `yhat1 +- qhat1` (naive CP) and `yhat1 <quantile> +- qhat1` (CQR PI), then:

- An user that used `show_all_PI = True` is not able to use the normal plotting method anymore as the normal plotting method only plots `yhat <quantile>` uncertainty intervals
- For someone starting with CQR, I think it's very confusing, but with the proposed naming you can easily see which one is the most current method and which one stems from original qr. 

But if you want those changes, although it would limit the use of the normal plotting method (only `conformal_plot` would be possible then), I can implement those.

leoniewgnr: done

LeonieFreisinger (approved):
LGTM!

Issue #1233, opened 2023-03-27
Major Frequency MS corresponds to 


Pushkaran-P:
Hi all,

- Environment Python 3.7, NeuralProphet version 0.5.3 installed from PYPI with pip install neuralprophet==0.5.3
- My data's major frequency isnt close to 100%, i would like to know why is this the case, this is the log I get when running NeuralProphet() "Major frequency MS corresponds to 89.091% of the data." 
- The data i'm working with
[sample.csv](https://github.com/ourownstory/neural_prophet/files/11097693/sample.csv)
-  Code for replication
 ```
test_horizon = 3
ID = datasample[datasample.columns[0]].reset_index(drop=True)[0]
row = datasample.values[0][1:]
firstnonzeropos = row.nonzero()[0][0]
valdf = pd.DataFrame(row[firstnonzeropos:last_col_index].astype(float))
timedf = times[firstnonzeropos:].reset_index(drop=True)
df = pd.concat([timedf,valdf],axis=1)
df.columns=['ds','y']
df_train = df[:-test_horizon]
m = NeuralProphet(weekly_seasonality=False,daily_seasonality=False, yearly_seasonality="auto",
                  n_changepoints=30, n_lags=6, n_forecasts = 12+test_horizon,
                 growth = "linear" )
metrics = m.fit(df_train, freq="MS",metrics=["MAE", "RMSE"])
print(metrics.tail(1))
```
<img width="752" alt="Screenshot 2023-03-27 124553" src="https://user-images.githubusercontent.com/84492623/227867978-dcc2b6ef-e812-407d-bdb6-1b0879468bc3.png">

```
future = m.make_future_dataframe(df_train, periods=m.n_forecasts)
forecast = m.predict(future)
```

<img width="563" alt="Screenshot 2023-03-27 124850" src="https://user-images.githubusercontent.com/84492623/227868593-f9187f3a-77cc-4e93-9630-e0203727d1d2.png">

Thanks in advance

noxan:
@Pushkaran-P thanks for providing feedback. I've tried the code snippet and dataset you posted, but it fails with `ValueError: could not convert string to float: 'v1'. Also when I try do dissect the code it seems to me that the loaded data might have the wrong format. Make sure to have one column with the time series information called `y` and one time related variable called `ds` (no additional columns are allowed).

Pushkaran-P:
[...]

PR #1264, opened 2023-04-14
[fix] Calculate major frequency percentage properly

noxan:
Fix for #1233

### DIFF
--- a/neuralprophet/df_utils.py
+++ b/neuralprophet/df_utils.py
@@ -1236,6 +1236,30 @@ def get_dist_considering_two_freqs(dist):
     return f1 + f2
 
 
+def _get_dominant_frequency_percentage(frequencies, distribution, filter_list) -> float:
+    """Calculate dominant frequency of dataframe.
+
+    Parameters
+    ----------
+        frequencies : list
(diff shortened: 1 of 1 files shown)


neuralprophet/df_utils.py
```diff
@@ -1236,6 +1236,29 @@ def get_dist_considering_two_freqs(dist):
...
+
+    Returns
+    -------
+        float
+            Percentage of dominant frequency within the whole dataframe
+
+    """
+    dominant_frequencies = [freq for freq in frequencies if freq in filter_list]
+    dominant_distribution = [distribution[np.where(frequencies == freq)] for freq in dominant_frequencies]
+    return sum(dominant_distribution) / sum(distribution)
```
LeonieFreisinger, line 1259: @noxan Can we ensure that we never divide by zero? I am not too deep into this function, but thought we should think abou it.

LeonieFreisinger (approved):
LGTM, nice job in improving readability

neuralprophet/df_utils.py
noxan: this would mean the dataset is empty

PR #1265, opened 2023-04-14
Revert "[minor] Allow lagged regressors with only unique values for global modeling"

leoniewgnr:
Reverts ourownstory/neural_prophet#1256


judussoari (approved):
Thanks for reverting!

PR #1266, opened 2023-04-14
[fix] Matplotlib and plotly plot showing aligned

leoniewgnr:
## :microscope: Background
To show the plots only once you had to use:
`m.plot(forecast)` for plotly and
`fig = m.plot(forecast)` for matplotlib

## :crystal_ball: Key changes
1. Now both plotting backends work with the same command.

  `m.plot(forecast)`
  `m.plot_parameters(forecast)`
  `m.plot_components(forecast)`
  is the correct command for plotly **and** matplotlib now

2. changed the new notebooks tutorial from `fig=m.plot(forecast)` to `m.plot(forecast)`


noxan (approved):
LGTM

tg2k:
@leoniewgnr I believe the `fig = fig.tight_layout()` changes such as https://github.com/ourownstory/neural_prophet/blame/5c45f4692463e3741c7389df8c056297f4a07cb5/neuralprophet/plot_forecast_matplotlib.py#L153 caused a regression.

Per https://github.com/matplotlib/matplotlib/blob/main/lib/matplotlib/figure.py#L3458 , I don't see `figure.tight_layout()` returning a value. In the debugger, this assignment changes `fig` to `None`.

A workaround is to fetch the figure after the fact with:

```
    from matplotlib import pyplot as plt
    fig = plt.figure(plt.get_fignums()[-1])    
```

PR #1267, opened 2023-04-14
[minor] Allow lagged regressors with only unique values for global modeling

judussoari:
## :microscope: Background

- Same as #1256 but now in accordance with refactored `forecaster.py`.
- Currently, NeuralProphet does not allow a lagged regressor to only contain unique values. When global modeling, however, there are reasonable cases in which some IDs may have only unique values, while others don't.

## :crystal_ball: Key changes

- Instead of throwing an Error immediately, we first check whether we have unique values cross all IDs. If not, we give a warning.


leoniewgnr (approved):
LGTM

PR #1250, opened 2023-04-07
[fix] Resolve "drop() got multiple values for argument 'axis'" error

noxan:
#1227

PR #1272, opened 2023-04-15
[fix] Remove regressor config if all regressors are removed due to unique values 

### LINKED ISSUE #1227: drop() got multiple values for argument 'axis'
### Discussed in https://github.com/ourownstory/neural_prophet/discussions/1220

<sup>Originally posted by **kzaho** March 16, 2023</sup>
F-n check_dataframe at (https://github.com/ourownstory/neural_prophet/blob/main/neuralprophet/df_utils.py) fails.
Is there an extra star symbol, which should be deleted?
```
regressors_to_remove = list(set(regressors_to_remove))
checked_df = checked_df.drop(*regressors_to_remove, axis=1)
```

judussoari:
## :microscope: Background

Fixes #1227 
Closes #1250 
In the case of only having unique values for future (or lagged) regressors, we currently remove the regressor column from the data. However, the `config_regressor` still survives (although being empty), which causes an error. 

## :crystal_ball: Key changes

In the event of all future (or lagged) regressors are removed due to only having single values in the column, the fitting and predicting procedure runs through without considering any regressors (i.e. the model is trained as if no regressors would have been specified). 

### DIFF
--- a/neuralprophet/data/process.py
+++ b/neuralprophet/data/process.py
@@ -299,7 +299,7 @@ def _check_dataframe(
             checked dataframe
     """
     df, _, _, _ = df_utils.prep_or_copy_df(df)
-    df, regressors_to_remove = df_utils.check_dataframe(
+    df, regressors_to_remove, lag_regressors_to_remove = df_utils.check_dataframe(
         df=df,
         check_y=check_y,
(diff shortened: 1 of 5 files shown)


noxan (approved):
Ah that makes sense, thanks for hunting down that bug @judussoari

PR #1280, opened 2023-04-18
[minor] Update dependencies to latest versions

noxan:
- Update the poetry lock file to use latest package versions


leoniewgnr (approved):
LGTM

PR #1282, opened 2023-04-18
[fix] Raise Exception again to avoid UnboundLocalError

### LINKED ISSUE #1084: Exception handling for too small dataset to detect learning rate is missleading
In case the dataset is too small to detect a learning rate the following warning is show:

`ERROR - (NP.utils.smooth_loss_and_suggest) - The number of loss values (1) is too small to estimate a learning rate. Increase the number of samples or manually set the learning rate.`

Yet the python exception thrown is very misleading with:

`UnboundLocalError: local variable 'suggestion' referenced before assignment`

Solution:

We should raise a `ValueError` or some other more suitable error message directly in `neuralprophet/utils.py:713` (inside `smooth_loss_and_suggest`) and make sure to not trigger an `UnboundLocalError`

noxan:
Fix #1084

### DIFF
--- a/neuralprophet/utils.py
+++ b/neuralprophet/utils.py
@@ -730,6 +730,7 @@ def smooth_loss_and_suggest(lr_finder_results, window=10):
         log.error(
             f"The number of loss values ({len(loss)}) is too small to estimate a learning rate. Increase the number of samples or manually set the learning rate."
         )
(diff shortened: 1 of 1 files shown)


leoniewgnr (approved):
LGTM

PR #1283, opened 2023-04-18
[minor] Upgrade holiday package to 0.23.0

noxan:
- Upgrade holidays package to 0.23.0
- Ensure we always use English for holiday names (else localization messes up merging multiple country holidays)


LeonieFreisinger (approved):
LGTM ! However, have another look into why the test is failing.

PR #1286, opened 2023-04-19
[minor] Add static plotly backend

### LINKED ISSUE #1235: Add plotly-static plotting backend

hxyue1:
## :microscope: Background

This mostly resolves #1235. I'm not sure how to proceed with plots for the conformal class since it has a different implementation.

## :crystal_ball: Key changes

Adds argument that allows users to select a static display type. If this is specified, it sets the backend to plotly and then displays the plot in the specified type. The interactive plot is still returned from the function call.


neuralprophet/plot_forecast_plotly.py
```diff
@@ -339,6 +347,8 @@ def plot_components(
     for ax in multiplicative_axes:
         ax = set_y_as_percent(ax)
     unregister_plotly_resampler()
+    if plotly_static:
```
LeonieFreisinger, line 350: @hxyue1 I am wondering, if in the 'plotly_static' case, we will display the figure twice if calling another .show() when applying the plot function?

LeonieFreisinger (changes requested):
Hey @hxyue1! Thanks a lot for your work in this PR. Looks generally good to me. :)
3 points to touch upon:
- if we call a `fig.show()` in the 'plotly_static' case and return the `fig`, do we run in a problem of plotting it twice with this function call. 
- can you edit the `test_plot()` in the t`est_plotting.py` with testing the 'plotly_static' ? I think it would be good the have it checked via at least one test. Editing all tests would be a little much. Would be great to also add the `plotly_resampler ` on that go
- Can as well adjust the plotly tutorial with the update of the `plotly_static` backend?

LeonieFreisinger (approved):
Great work! Thanks for keeping up with it

LeonieFreisinger:
@hxyue1 please have a final look at resolving the merge conflict. Afterwards, feel free to merge.

PR #1298, opened 2023-04-21
[fix] Remove SmoothL1Loss from Tutorial

### LINKED ISSUE #1234: cross-validation demo error
When I was studying the cross-validation page in the official document, I found that the following code would report an error, prompting "['SmoothL1Loss'] not in index". The output of the test metrics does not include these items. Is it because the version update has been modified?
```
metrics_train = pd.DataFrame(columns=METRICS)
metrics_test = pd.DataFrame(columns=METRICS)

for df_train, df_test in folds:
    m = NeuralProphet(**params)
    train = m.fit(df=df_train, freq="MS")
    test = m.test(df=df_test)
    metrics_train = metrics_train.append(train[METRICS].iloc[-1])
    metrics_test = metrics_test.append(test[METRICS].iloc[-1]
```
test metrics like this
image

judussoari:
## :microscope: Background

Closes #1234 
SmoothL1Loss is not supported by Torchmetrics. With the lightning migration, we only use those, however. 
Also, `m.test()` does not return error metrics specified by the user. 

## :crystal_ball: Key changes

- Remove SmoothL1Loss in crossvalidation tutorial. 
- The return values of m.test() have been adjusted. 


neuralprophet/time_net.py
```diff
@@ -793,6 +793,9 @@ def test_step(self, batch, batch_idx):
         loss, reg_loss = self.loss_func(inputs, predicted, targets)
         # Metrics
         if self.metrics_enabled:
+            predicted_denorm = self.denormalize(predicted[:, :, 0])
+            target_denorm = self.denormalize(targets.squeeze(dim=2))
+            self.log_dict(self.metrics_val(predicted_denorm, target_denorm), **self.log_args)
             self.log("Loss_test", loss, **self.log_args)
```
judussoari, line 799: This is to log (and eventually return) the error metrics specified by the user. those 3 lines are in validation_step and training_step, but (apparently) have been forgottten here

leoniewgnr (approved):
LGTM

PR #1300, opened 2023-04-21
[fix] Raise ValueError when df contains not enough rows

### LINKED ISSUE #1294: Unclear error when doing auto-regression with too small dataset
### Discussed in https://github.com/ourownstory/neural_prophet/discussions/1257

<sup>Originally posted by **pzaika** April 11, 2023</sup>
When using auto-regression with a dataset that is smaller than the requested number of forecasts + lags (specifically when n_samples = len(df) - max_lags + 1 - n_forecasts is not positive) the fit method throws an IndexError. This hides the real problem, which is that the user is trying to train a model on not enough data. I think that it would be better to raise an error which tells the user to lower the number of forecasts or lags they are asking for.</div>

judussoari:
## :microscope: Background

- Fixes #1294 
- Closes #1257 

When using auto-regression with a dataset that is smaller than `n_forecasts` + `n_lags` the fit method throws an IndexError. This hides the real problem, which is that the user is trying to train a model on not enough data. 

## :crystal_ball: Key changes

- raise an error which tells the user to lower the number of forecasts and/or lags, or asks for a larger dataset

### DIFF
--- a/neuralprophet/data/process.py
+++ b/neuralprophet/data/process.py
@@ -397,6 +397,11 @@ def _check_dataframe(
         pd.DataFrame
             checked dataframe
     """
+    if len(df) < (model.n_forecasts + model.n_lags):
+        raise ValueError(
+            "Dataframe has less than n_forecasts + n_lags rows. "
+            "Forecasting not possible. Please either use a larger dataset, or adjust the model parameters."
+        )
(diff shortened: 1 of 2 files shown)


leoniewgnr (approved):
LGTM

PR #1311, opened 2023-04-28
[major] speed up nested additive model

LeonieFreisinger:
## :microscope: Background

The first implementation includes two forward() passes and hence unnecessary many function calls. --> Objective: Reduce computation time. 


ourownstory (approved):
Great work!! Amazing!!

We can single-track the compute_components later.

PR #1315, opened 2023-04-29
[devops] Fix metrics Github action

noxan:
## :microscope: Background

- The Github action metrics fail, this should fix it


noxan:
@leoniewgnr @LeonieFreisinger seems the metrics work again on main, should we close this one here then?

PR #1323, opened 2023-05-02
[devops] change workflow python version to 3.11

LeonieFreisinger:


noxan (changes requested):
Changes look good but our framework is not compatible with 3.11 yet

leoniewgnr:
**Python 3.11 Update**
I suggest we continue to support Python 3.10 for now. PyTorch 2.0, which this package depends on, currently lacks Conda binaries for Python 3.11 due to unmet dependencies like NumPy and SciPy on Anaconda. While Pip support for Python 3.11 is available, maintaining compatibility for both Pip and Conda users is advisable. We can reconsider Python 3.11 support once the dependencies and PyTorch Conda binaries for this version become available.

leoniewgnr:
**Google Colab Update**
The default version for google is 3.10, so I think we can definetely switch to 3.9 instead of 3.8

leoniewgnr:
Closing this issue with the following conclusion, can be opened again later.

Currently, we support from 3.8 to 3.10.

**Increase upper bound to 3.11?**
I suggest we continue to support Python 3.10 for now.[ PyTorch 2.0, which this package depends on, currently lacks Conda binaries for Python 3.11 due to unmet dependencies like NumPy and SciPy on Anaconda.](https://pytorch.org/blog/pytorch-2.0-release/#python-311-support-on-anaconda-platform) While Pip support for Python 3.11 is available, maintaining compatibility for both Pip and Conda users is advisable. We can reconsider Python 3.11 support once the dependencies and PyTorch Conda binaries for this version become available.
**--> IMO: No**

**Increase lower bound to 3.9?**
We decided to support for 3.8 because google colab used to use that. The current default for google colab is 3.10. So, theoretically we could increase to 3.9 but then we only support 3.9 and 3.10, so I suggest we increase to 3.9 only, when we also increased to 3.10
**--> IMO: No**

**Conclusion**: We can aim to keep supporting python versions that are still 'alive' - 3.8 will still be bug-fix supported for a year. Especially as the newer versions seem to get such slow support - 3.11 still not properly supported - with 3.12 releasing now...\

nrcjea001:
I believe Python 3.11 support is now available for Pytorch 2.1 ([see here](https://github.com/pytorch/pytorch/issues/86566#issuecomment-1706075651)), does this mean the upper bound to 3.11 is feasible?

PR #1324, opened 2023-05-10
[devops] Add ruff github action

noxan:
## :microscope: Background

- Ruff is a new super fast linting tool (could replace flake8 and isort) https://github.com/charliermarsh/ruff

### DIFF
--- a/.github/workflows/linters.yml
+++ b/.github/workflows/linters.yml
@@ -30,6 +30,14 @@ jobs:
           linters: flake8
           run: flake8 neuralprophet scripts
 
+  ruff:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: ruff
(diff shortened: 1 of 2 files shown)


leoniewgnr:
@noxan why does the linters /ruff test fail?

ourownstory (approved):
Great upgrade, thank you Richard!

What do you suggest to do about it's list of errors? Should we address all or should we relax what constitutes an error?

ourownstory:
@noxan Might it make sense to also add this to the pre-commit or pre-push hook?

noxan:
We should migrate our pre-commit hooks to https://pre-commit.com/ - great idea to include it there too.

We might be able to drop flake8 and isort over time (to be replaced by ruff).

Regarding the warnings: So far we excluded the `tests` folder. Ruff also has an autofix feature, I've pushed things to a branch here: https://github.com/ourownstory/neural_prophet/compare/ruff-fixes?expand=1

PR #1325, opened 2023-05-10
[devops] Remove macos and windows from Github actions

### LINKED ISSUE #991: Windows tests in github actions fail

noxan:
## :microscope: Background

- There has not been any difference between the operating systems so far

## :crystal_ball: Key changes

- Remove macos and windows, as linux runs fastest
- Fixed #991

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
     runs-on: ${{matrix.os}}
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-latest]
         python-version: ["3.8", "3.10"]
     env:
       POETRY_VIRTUALENVS_CREATE: false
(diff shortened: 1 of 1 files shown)


leoniewgnr (dismissed):
we decided only to remove macos

ourownstory (approved):
Makes sense, thank you!
Until we remove matplotlib support (or potentially longer), I suggest we keep windows test to catch any plotting package dependency/build issues.

noxan:
Windows tests fail frequently due to "This probably means that Tcl wasn't installed properly."

Also why should there be any different dependency or build issues on windows?

PR #1332, opened 2023-05-18
[docs] Fix prophet tutorial

### LINKED ISSUE #939: Migration from prophet jupyter notebook does not build for docs

leoniewgnr:
## :microscope: Background
- jupyter notebook fails while building docs
- fixes #939 

## :crystal_ball: Key changes
- Workaround: code in markdown cell


LeonieFreisinger (approved):
LGTM

ourownstory (approved):
Good fixes, thank you!

I corrected the now added support for Conditional seasonality to be included. 

Further, I think we can drop 
https://github.com/ourownstory/neural_prophet/blob/fix-prophet-tutorial/docs/source/how-to-guides/changes-from-prophet.md
Do you mind quickly checking if it's contents are contained within this guide?
(feel free to merge this and remove it in a separate PR)

Finally, I remember we also had a tutorial for moving from prophet to NeuralProphet (not just TorchProphet - I could not find it quickly again. However, it is the more important of the two, and we should revive it if anyhow possible.

Issue #1346, opened 2023-06-12
prediction_frequency not working when quantiles are set


leoniewgnr:
**Describe the bug**
When `quantiles` are set, `prediction_frequency` cannot be set as `components[comp]` in `process.py`, line 160 does not match from the size. 

**To Reproduce**
`params = {
    'n_lags': 7*24, 
    'n_forecasts': 33,
    'prediction_frequency': {'daily-hour': 14},
    'quantiles': quantiles,
}

m = NeuralProphet(**params)

metrics = m.fit(df)

forecast = m.predict(df)`

leoniewgnr:
solved with #1347

leoniewgnr:
solved with #1347

PR #1347, opened 2023-06-12
[minor] fixed prediction_frequency when quantiles are set

leoniewgnr:
## :microscope: Background
see #1346 

## :crystal_ball: Key changes
only used first dimension of `components[comp]`


leoniewgnr:
can anybody confirm that the first dimension of component[comp] corresponds to the 50% forecast?

LeonieFreisinger (approved):
LGTM!

Issue #1348, opened 2023-06-12
lagged regressors are not added to yhat forecast


leoniewgnr:
**Describe the bug**
yhat is only summed up of ar, trend and seasonalities but not lagged regressors

**To Reproduce**
# Model and prediction
m = NeuralProphet(
    epochs=10,
    n_lags=12,
    daily_seasonality=False,
    yearly_seasonality=False,
    weekly_seasonality=False,
)

# Add the new lagged regressor
m.add_lagged_regressor("temperature")

PR #1350, opened 2023-06-13
[minor] fix sum of components in 0.6.0

leoniewgnr:
## :microscope: Background
see #1348 

## :crystal_ball: Key changes
call `compute_components` with `corrected_input`

## 🚨 Attention
This PR is only for release 0.6.0.
It cannot be merged on the current main as the code changed since then


ourownstory:
@leoniewgnr Thank you for creating this for 0.6.0 too. 
We need to however change the merge target to a branch `release/0.6.1` that was forked off main at the release 0.6.0 state. Would you mind creating such a branch?

PR #1357, opened 2023-06-21
Update CONTRIBUTING.md Poetry dev install

ourownstory:
The dev install instructions are still for pip and setuptools, but we have moved to poetry. This PR is for updating the instructions.


ourownstory:
@hxyue1 Here is a beginning of a draft for updating the dev wiki.

PR #1360, opened 2023-06-27
[major] Speed up fit function

leoniewgnr:
## :microscope: Background
We all know the fit function is super slow. This is due to some bottlenecks that are responsible for most of the slow code.

 
## :crystal_ball: Key changes
Speed up the following bottlenecks

- [ ] 'init_after_tabularized' in time_dataset.py
- [ ] first for loop 'drop_nan_after_init' in time_dataset.py


LeonieFreisinger (approved):
LGTM

Issue #1361, opened 2023-06-27
encountered different devices in metric calculation 


webcoderz:
**Prerequisites**

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in a new virtualenv with only neuralprophet installed, directly from github:
      ```shell
      git clone <copied link from github>
      cd neural_prophet
      pip install .
      ```
    * Checked the Answered Questions on the Github Disscussion board: https://github.com/ourownstory/neural_prophet/discussions
      If you have the same question but the Answer does not solve your issue, please continue the conversation there.
    * Checked that your issue isn't already filed: https://github.com/ourownstory/neural_prophet/issues
      If you have the same issue but there is a twist to your situation, please add an explanation there.
    * Considered whether your bug might actually be solveable by getting a question answered: 
      * Please [post a package use question](https://github.com/ourownstory/neural_prophet/discussions/categories/q-a-get-help-using-neuralprophet)
      * Please [post a forecasting best practice question](https://github.com/ourownstory/neural_prophet/discussions/categories/q-a-forecasting-best-practices)
      * Please [post an idea or feedback](https://github.com/ourownstory/neural_prophet/discussions/categories/ideas-feedback)

**Describe the bug**

when doing multi threaded gpu training sessions the metrics are not passing the tensors from cpu to gpu and am receiving this error:
`Encountered different devices in metric calculation . This could be due to the metrics class not being on same device as the input. Instead of MeanAbsoluteError() use MeanAbsoluteError().to(device)`

**To Reproduce**

1. run multi threaded training session with gpu enabled in the trainer config

**Expected behavior**

For metrics to cleanly pass from cpu into GPU. the utils_metric.py only calls the error function to be used ([L7](https://github.com/ourownstory/neural_prophet/blob/2b303f8ddd2859e050c4492dd8c983a1fcada76f/neuralprophet/utils_metrics.py#L7)), no sort of device tracking . 

**What actually happens**

Describe what happens, and how often it happens.

**Screenshots**

If applicable, add screenshots and console printouts to help explain your problem.

**Environement (please complete the following information):**

 - Python environment 3.8.16
 - NeuralProphet version and install method [e.g. 2.7, installed from PYPI with `pip install neuralprophet`] 0.5.4

**Additional context**

leoniewgnr:
Do you have some piece of minimal code for me? Then I'd be happy to fix the bug:)

webcoderz:
[...]

PR #1365, opened 2023-07-04
[bug] Fix correct definition of torchmetrics inside pytorch lightning module

leoniewgnr:
## :microscope: Background
See #1361 

## :crystal_ball: Key changes
Move metrics definition inside pytorch lightning module as described in https://torchmetrics.readthedocs.io/en/latest/pages/lightning.html


neuralprophet/time_net.py
```diff
@@ -163,8 +163,17 @@ def __init__(
         self.batch_size = self.config_train.batch_size
 
         # Metrics Config
+        METRICS = {
```
ourownstory, line 166: We currently have `METRICS` defined in here and `utils_metrics.py`. We should only have one source for this dict. Can we consolidate this?

neuralprophet/utils_metrics.py
```diff
@@ -22,21 +22,21 @@ def get_metrics(metric_input):
 
     Returns
     -------
-        dict
-            Dict of torchmetrics.Metric metrics.
+        dict or list
```
ourownstory, line 25: Maybe we should rename this function to `check_metrics`

neuralprophet/utils_metrics.py
```diff
@@ -22,21 +22,21 @@ def get_metrics(metric_input):
...
+            return [metric_input]
         else:
             raise ValueError("Received unsupported argument for collect_metrics.")
     elif isinstance(metric_input, list):
         if all([m.upper() in METRICS.keys() for m in metric_input]):
-            return {k: v for k, v in METRICS.items() if k in metric_input}
+            return metric_input
         else:
             raise ValueError("Received unsupported argument for collect_metrics.")
     elif isinstance(metric_input, dict):
```
ourownstory, line 42: These user specified metrics will be defined outside and thus not be associated with a device. Might this lead to issues?

neuralprophet/time_net.py
```diff
@@ -163,8 +163,17 @@ def __init__(
...
+            "MAE": torchmetrics.MeanAbsoluteError(),
+            "MSE": torchmetrics.MeanSquaredError(squared=True),
+            "RMSE": torchmetrics.MeanSquaredError(squared=False),
+        }
+
         self.metrics_enabled = bool(metrics)  # yields True if metrics is not an empty dictionary
         if self.metrics_enabled:
+            # only convert to dict if metrics is a list (metrics were not set in utils_metrics)
+            if isinstance(metrics, list):
+                metrics = {metric: METRICS[metric] for metric in metrics}
```
ourownstory, line 176: Does moving `METRICS` to this file and transfering the objects to this dict resolve the device mismatch issue?

ourownstory (changes requested):
Looking good mostly!
We just need make sure to only have one source for the list of `METRICS`. 
I also added some other open Qs

neuralprophet/time_net.py
leoniewgnr: yes it works now

neuralprophet/utils_metrics.py
leoniewgnr: Resolved in newest commit

neuralprophet/time_net.py
leoniewgnr: only defined in utils_metrics.py

neuralprophet/utils_metrics.py
leoniewgnr: function changed, now we are still getting metrics (only metric names, but still getting the metrics)

hxyue1 (approved):

Issue #1229, opened 2023-03-22
Transition to Pytorch 2.0


ourownstory:

noxan:
It is a bit tricky to get a good overview as one issue leads to another.
We definitely have to adjust for the trainer and auto tuner separation
- https://github.com/Lightning-AI/lightning/pull/16462

hxyue1:
Using `poetry install` with `torch = "^2.0.0"` will by default install the gpu version.  For machines which do not have a gpu available, this will lead to an immediate crash upon importing `torch`

```
Python 3.8.16 (default, Jun 12 2023, 18:09:05) 

[GCC 11.2.0] :: Anaconda, Inc. on linux

Type "help", "copyright", "credits" or "license" for more information.

>>> import torch

Traceback (most recent call last):

  File "/home/hxyue1/anaconda3/envs/torch2/lib/python3.8/site-packages/torch/__init__.py", line 168, in _load_global_deps

    ctypes.CDLL(lib_path, mode=ctypes.RTLD_GLOBAL)

  File "/home/hxyue1/anaconda3/envs/torch2/lib/python3.8/ctypes/__init__.py", line 373, in __init__

    self._handle = _dlopen(self._name, mode)

OSError: libcurand.so.10: cannot open shared object file: No such file or directory



During handling of the above exception, another exception occurred:



Traceback (most recent call last):

  File "<stdin>", line 1, in <module>

  File "/home/hxyue1/anaconda3/envs/torch2/lib/python3.8/site-packages/torch/__init__.py", line 228, in <module>

    _load_global_deps()

  File "/home/hxyue1/anaconda3/envs/torch2/lib/python3.8/site-packages/torch/__init__.py", line 189, in _load_global_deps

    _preload_cuda_deps(lib_folder, lib_name)

  File "/home/hxyue1/anaconda3/envs/torch2/lib/python3.8/site-packages/torch/__init__.py", line 154, in _preload_cuda_deps

    raise ValueError(f"{lib_name} not found in the system path {sys.path}")

[...]

PR #1373, opened 2023-07-10
Torch2

hxyue1:
## :microscope: Background

- This is to upgrade neuralprophet to use pytorch 2.0 and address #1229

## :crystal_ball: Key changes

- Update the pyproject.toml file to use pytorch 2.0 and add optional CPU dependencies.

- [] I have commented my code, added docstrings and data types to function definitions.
- [] I have added pytests to check whether my feature / fix works.


.github/workflows/ci.yml
```diff
@@ -31,8 +31,12 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install Poetry
         uses: snok/install-poetry@v1
+      - name: Generate lock file
+        run: poetry lock
+      - name: Update lock file
+        run: poetry update
```
noxan, line 37:
```suggestion
```
The lockfile should not be automatically updated by our CI job, but rather locally - else we lose the benefits of locked dependencies from poetry.

Best update the lockfile locally - without updating any of the other dependencies with `poetry lock --no-update`

.github/workflows/ci.yml
```diff
@@ -56,6 +60,10 @@ jobs:
         uses: abatilo/actions-poetry@v2
       - name: Setup Pandoc
         uses: r-lib/actions/setup-pandoc@v2
+      - name: Generate lock file
+        run: poetry lock
+      - name: Update lock file
+        run: poetry update
```
noxan, line 66:
```suggestion
```

see above

.github/workflows/linters.yml
```diff
@@ -84,6 +84,10 @@ jobs:
       - name: Setup poetry
         if: steps.poetry-cache.outputs.cache-hit != 'true'
         uses: snok/install-poetry@v1
+      - name: Generate lock file
+        run: poetry lock
+      - name: Update lock file
+        run: poetry update
```
noxan, line 90:
```suggestion
```
see above

pyproject.toml
```diff
@@ -29,16 +29,28 @@ kaleido = "0.2.1"
 plotly-resampler = "^0.8.3.1"
 pytorch-lightning = "^1.9.4"
 tensorboard = "^2.11.2"
-torch = "^1.13.1"
+torch = "^2.0.0"
```
noxan, line 32:
```suggestion
torch = "^2.0.1"
```
there is torch 2.0.1 available already

noxan (changes requested):

hxyue1:
Unfortunately this fails on windows with python 3.8 for some reason, but is fine in all other scenarios. The flake8 and pyright jobs are also failing but I haven't made changes to any python code so I'm not sure why that's the case.

noxan:
@hxyue1, thanks for the update and your work. To better understand the different problems you encountered, I drafted #1404.

I am not a big fan of forcing the `cpu` version of Pytorch - see my comment https://github.com/ourownstory/neural_prophet/issues/1229#issuecomment-1681393452 - let's come up with a plan there and then get one of the pull requests merged.

leoniewgnr:
@hxyue1 I will close this for now because #1404 fixes it, but feel free to open it again if you need it.

PR #1400, opened 2023-08-14
[minor] Memory optimized stride function

leoniewgnr:
## :microscope: Background
def _stride_time_features_for_forecasts(x) uses a lot of memory

## :crystal_ball: Key changes
replace with `np.lib.stride_tricks.as_strided`


noxan (approved):
Great improvement on readability and performance 👍

PR #1402, opened 2023-08-15
[minor] Memory efficient float32 types instead of float64 types

leoniewgnr:
## :microscope: Background
The whole computation was done with float64 types, but torch tensors are float32 anyway and most of the ML models use float32 values. This will affect model accuracy but I think its worth it because we safe a lot of memory!

## :crystal_ball: Key changes

- y and yhat values: float32 if float32 is inserted, float64 if float64 is inserted
- time values: will always be float64 because `np.datetime64`
- feature values (events): float32

### DIFF
--- a/neuralprophet/data/process.py
+++ b/neuralprophet/data/process.py
@@ -132,7 +132,7 @@ def _reshape_raw_predictions_to_forecst_df(
             for j in range(len(quantiles)):
                 forecast_0 = components[comp][0, :, j]
                 forecast_rest = components[comp][1:, n_forecasts - 1, j]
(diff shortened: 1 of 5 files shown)


neuralprophet/data/process.py
leoniewgnr, line 1: If user inputs float64, it will be converted to float32

neuralprophet/df_utils.py
leoniewgnr, line 1: This line is unnecessary as they will be converted to datetime the line below

noxan (approved):
LGTM

PR #1408, opened 2023-08-24
[bug] fixed ruff warnings

leoniewgnr:
## :microscope: Background
ruf warnings because type instead of isinstance was used


LeonieFreisinger (approved):
LGTM

Issue #1415, opened 2023-09-07
Size of saved model file changes dramatically after different calls of `predict` 


c3-ziqin:
**Describe the bug**

Time and disk space needed for saving a model changes dramatically after different calls of `predict`. Potentially this can be expected, as some info about the forecasted dataset might be saved on `NeuralProphet` object. 

However, giving users an option to only save necessary part of model (enough for future predictions) can be a huge plus.  

**To Reproduce**

#### save the model directly after training
```python
import pandas as pd
from neuralprophet import NeuralProphet, set_log_level, df_utils
from neuralprophet import save as npsave

data_location = "https://raw.githubusercontent.com/ourownstory/neuralprophet-data/main/datasets/yosemite_temps.csv"
df = pd.read_csv(data_location)

m = NeuralProphet()
_ = m.fit(df, epochs=2, learning_rate=0.01, progress=None)

# save the model directly after training
fn = "test1.np"
%time npsave(m, fn)
print(f"np file size in MB: {os.path.getsize(fn) / 1e6}")
```
```
CPU times: user 3.86 s, sys: 29.8 ms, total: 3.89 s
Wall time: 3.9 s
np file size in MB: 13.682805
```
#### save model after predict on 2 rows
```python
m = NeuralProphet()
_ = m.fit(df, epochs=2, learning_rate=0.01, progress=None)
# run forecast with 2 rows
forecast = m.predict(df.iloc[-2:])
# save model after predict on 2 rows
fn = "test2.np"
%time npsave(m, fn)
print(f"np file size in MB: {os.path.getsize(fn) / 1e6}")
```
```
CPU times: user 4.89 ms, sys: 602 µs, total: 5.5 ms
Wall time: 5.56 ms
np file size in MB: 0.045449
```
#### save model after predict on 1000 rows
```python
m = NeuralProphet()
_ = m.fit(df, epochs=2, learning_rate=0.01, progress=None)
# run forecast with 1000 rows
forecast = m.predict(df.iloc[-1000:])
# save model after predict on 1000 rows
fn = "test3.np"
%time npsave(m, fn)
print(f"np file size in MB: {os.path.getsize(fn) / 1e6}")
```
```
CPU times: user 172 ms, sys: 1.9 ms, total: 174 ms
Wall time: 174 ms
np file size in MB: 0.761993
```

**Expected behavior**
- Similar size of model files should be generated, OR
- adding an option in `save` function to only save the minimal model

**Additional context**
[...]

PR #1425, opened 2023-09-19
[bug] Fix saved model size

leoniewgnr:
## :microscope: Background
See #1415 

## :crystal_ball: Key changes
removed attribute `_trainer`


leoniewgnr:
I removed some attributes, but really unsure where the data is saved. In `model.logger.history` is saved some big tensor, but the model size is still not the same:
```
2023-09-19 13:04:06 [INFO]: size of saved fitted model: 288791 (test_utils.py:41)
2023-09-19 13:04:06 [INFO]: testing: save forecast model (test_utils.py:45)
2023-09-19 13:04:06 [INFO]: size of saved forecast model: 43850 (test_utils.py:47)
```

leoniewgnr:
I think I found it! In _trainer there was saved some unecessary information!
2023-09-19 13:51:06 [INFO]: size of saved fitted model: 18095 (test_utils.py:43)
2023-09-19 13:51:06 [INFO]: testing: save forecast model (test_utils.py:47)
2023-09-19 13:51:06 [INFO]: size of saved forecast model: 18180 (test_utils.py:50)

ourownstory (approved):
Good fix! 

Do you think we should just include `"_trainer"` in  `attrs_to_remove = ["trainer"]`?
Or is there a relevant scenario where  `"_trainer"` needs to be saved, but  `"trainer"`  not?
If not, we can skip the 'minimal' option and keep the code more lean.

leoniewgnr:
@ourownstory no probably not, I think so too, we should drop the minimal version and keep the code lean, will change that soon

PR #1439, opened 2023-10-04
[devops] Update poetry lock to most recent dependencies

ourownstory:
## :microscope: Background

- As a regular pip install will not include the lock file dependencies, we need to update the lock file regularly or omit it. At first we are attempting a regular update. 

## :crystal_ball: Key changes

- The first full reset to the newest dependencies is contained in this update.


ourownstory:
merging despite the one failed windows test, as it is needs a separate fix.
`tests\test_plotting.py::test_plot_conformal_prediction_asymmetric[matplotlib] - _tkinter.TclError`

PR #1509, opened 2024-01-16
Fix - loading model on machine with diffrent accelerator

McOffsky:
## :microscope: Background

During my experiments with NP I've found out, that utils.load() is using trainer config is retrived from model, including accelerator. This caused an error on cpu machine while loading model fitted on GPU.

## :crystal_ball: Key changes

- NeuralProphet.restore_trainer() now takes optional argument, accelerator: str = None. If not provided, value stored in model will be used.

- utils.load() now creates torch.device(map_location) into separate variable, and that variable is used in torch.load(). Orginal value is additionaly passed to NeuralProphet.restore_trainer() as accelerator

- [n/c] I have added pytests to check whether my feature / fix works.


ourownstory (approved):
Great addition, thank you!

ourownstory:
Can be merged after passing of tests.

McOffsky:
Happy to help with such great library! 
I'll fix the issues tomorrow, it'a a bit late in my timezone ;)

ourownstory:
Hi @McOffsky Thank you!
The failed tests are not related to your contribution, we can safely ignore them. 

The pyright type checker did want a change of line 2737, I fixed it quickly - save your efforts for a later contribution. 
(For your reference you can read about the type logic here: https://stackoverflow.com/questions/51710037/how-should-i-use-the-optional-type-hint)

Best, 
Oskar

ourownstory:
Update: waiting for rerun of tests to finish before merge.

PR #1514, opened 2024-01-17
[devops] Upgrade to lightning 2.0

ourownstory:
Upgrade guide
https://lightning.ai/docs/pytorch/latest/upgrade/from_1_9.html


ourownstory (commented):
It seems like the tests are yielding differing metrics - Do you have any idea what might be the cause?

pyproject.toml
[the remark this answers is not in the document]
MaiBe-ctrl: I haven't changed that requirement. I'll add it shortly.

pyproject.toml
```diff
@@ -16,28 +16,27 @@ classifiers = [
 Homepage = "https://github.com/ourownstory/neural_prophet"
 
 [tool.poetry.dependencies]
-python = "^3.9"
-typing-extensions = "^4.5.0"
+python = ">=3.9,<3.12"
```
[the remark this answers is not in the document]
MaiBe-ctrl: I haven't changed that requirement. I'll add it shortly.

MaiBe-ctrl:
@ourownstory Not sure where the differences in the metrics are coming from, but I had to cast some variables to float32 to fix some tests which might caused these differences.

neuralprophet/configure.py
```diff
@@ -311,12 +311,12 @@ def __post_init__(self):
             self.trend_local_reg = False
 
         # If trend_local_reg = True
-        if self.trend_local_reg == True:
+        if self.trend_local_reg:
```
ourownstory, line 314: This change looks wrong - see comment in other PR

neuralprophet/configure.py
```diff
@@ -398,12 +398,12 @@ def __post_init__(self):
             self.seasonality_local_reg = False
 
         # If seasonality_local_reg = True
-        if self.seasonality_local_reg == True:
+        if self.seasonality_local_reg:
```
ourownstory, line 401: This also overwrites existing config, which should not happen

ourownstory (commented):
Looks like two changes to the if-check have unintentional consequences. @MaiBe-ctrl

neuralprophet/configure.py
```diff
@@ -311,12 +311,12 @@ def __post_init__(self):
...
 
         # If trend_local_reg = True
-        if self.trend_local_reg:
+        if self.trend_local_reg is True:
             log.error("trend_local_reg = True. Default trend_local_reg value set to 1")
             self.trend_local_reg = 1
 
         # If Trend modelling is global.
-        if self.trend_global_local == "global" and self.trend_local_reg:
+        if self.trend_global_local == "global" and self.trend_local_reg is True:
```
ourownstory, line 319: Here,   your  change was actually correct before:
` if self.trend_global_local == "global" and self.trend_local_reg:`

neuralprophet/configure.py
```diff
@@ -398,12 +398,12 @@ def __post_init__(self):
...
 
         # If seasonality_local_reg = True
-        if self.seasonality_local_reg:
+        if self.seasonality_local_reg is True:
             log.error("seasonality_local_reg = True. Default seasonality_local_reg value set to 1")
             self.seasonality_local_reg = 1
 
         # If Season modelling is global.
-        if self.global_local == "global" and self.seasonality_local_reg:
+        if self.global_local == "global" and self.seasonality_local_reg is True:
```
ourownstory, line 406: Here,   your  change was actually correct before:
`        if self.global_local == "global" and self.seasonality_local_reg:`

ourownstory:
> @ourownstory Not sure where the differences in the metrics are coming from, but I had to cast some variables to float32 to fix some tests which might caused these differences.

I think the difference comes from the potentially changed implementation of the learning rate finder, that now is part of lightning Tuner.

Issue #1577, opened 2024-06-06
Adding auto regression n_lags- seems to remove any predictions or cause a length mismatch in datasets? very basic- 


suntochukwu:
I feel like this is extremely basic. The data set is just gold stock market prices- ds/datetime and closing price/y. 

```
mgc_data = yf.download('MGC=F', period='1y', interval='1h')
mgc_data= pd.DataFrame(mgc_data).reset_index()
mgc_data.rename(columns={'Datetime': 'ds' ,'Close': 'y' }, inplace=True)


#--prophet
basicdf= mgc_data[['ds','y']]
basicdf.dropna(inplace=True)`
```

Here is the code:

```
m2 = NeuralProphet(drop_missing= True , growth= 'discontinuous', n_lags=50)
m2.set_plotting_backend("plotly-static")
metrics = m2.fit(basicdf, freq= 'H')
#future = m2.make_future_dataframe(basicdf, periods=50, n_historic_predictions=True)
forecast = m2.predict(future) 
```

Without n_lags the models works and predicts well. When i add autoregression i get an error:

dataset returned: 
	ds	step0	trend0	season_weekly0	season_daily0	ar0	step1	trend1	season_weekly1	season_daily1	...	step98	trend98	season_weekly98	season_daily98	ar98	step99	trend99	season_weekly99	season_daily99	ar99
0	2023-06-08 20:00:00	NaN	NaN	NaN	NaN	NaN	NaN	NaN	NaN	NaN	...	NaN	NaN	NaN	NaN	NaN	NaN	NaN	NaN	NaN	NaN
1	2023-06-08 21:00:00	NaN	NaN	NaN	NaN	NaN	NaN	NaN	NaN

PR #1585, opened 2024-06-17
[Minor] fix future regressor

### LINKED ISSUE #1567: null prediction when using future regressor and len(training set) == n_lags + n_forecasts
**Prerequisites**

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in a new virtualenv with only neuralprophet installed, directly from github:
      ```shell
      git clone <copied link from github>
      cd neural_prophet
      pip install .
      ```
    * Checked the Answered Questions on the Github Discussion board: https://github.com/ourownstory/neural_prophet/discussions
      If you have the same question but the Answer does not solve your issue, please continue the conversation there.
    * Checked that your issue isn't already filed: https://github.com/ourownstory/neural_prophet/issues
      If you have the same issue but there is a twist to your situation, please add an explanation there.
    * Considered whether your bug might actually be solvable by getting a question answered: 
      * Please [post a package use question](https://github.com/ourownstory/neural_prophet/discussions/categories/q-a-get-help-using-neuralprophet)
      * Please [post a forecasting best practice question](https://github.com/ourownstory/neural_prophet/discussions/categories/q-a-forecasting-best-practices)
      * Please [post an idea or feedback](https://github.com/ourownstory/neural_prophet/discussions/categories/ideas-feedback)

**Describe the bug**

When using future regressor, and the training dataset length is exaclty equal to n_lags + n_forecasts, the model.predict methods returns a prediction dataframe with null values on all yhat columns. 

The same training data works when future regressor is not used.

Please run the included code to reproduce the error.

**To Reproduce**

```
import pandas as pd
import numpy as np
import neuralprophet
demo_df = pd.DataFrame(
   {
    "ds": pd.date_range("2022-04-01", "2022-04-30"),
   "y": np.random.random(30),
   "reg": np.random.random(30),
   })

demo_model = neuralprophet.NeuralProphet(n_forecasts=15, n_lags=15)
demo_model.set_plotting_backend('matplotlib')
demo_model.add_future_regressor("reg")
demo_model.fit(demo_df)
prediction = demo_model.predict(demo_df) #<== THIS IS THE BLANK DATAFRAME
demo_model.plot(prediction) 


demo_model2 = neuralprophet.NeuralProphet(n_forecasts=15, n_lags=15)
demo_model2.set_plotting_backend('matplotlib')
# demo_model.add_future_regressor("reg")
demo_model2.fit(demo_df[['ds','y']])
prediction2 = demo_model2.predict(demo_df[['ds','y']])  #<== THIS IS THE EXPECTED BEAHVIOUR
demo_model2.plot(prediction2)
```

**Expected behavior**
The model.predict method should return a dataframe with predicted values for the n_forecasts dates

**What actually happens**
The model.predict method returns a dataframe with null values in all yhat columns

**Environment (please complete the following information):**

 - Python environment Python: 3.11.9, venv
 - NeuralProphet version and install method: v0.8.0 installed from PYPI with `pip install neuralprophet`

MaiBe-ctrl:
## :microscope: Background

- Closes #1567 #1577

## :crystal_ball: Key changes

- Explain the main changes introduced by this pull request for the reviewer.


neuralprophet/forecaster.py
```diff
@@ -1107,7 +1112,7 @@ def fit(
         self.fitted = True
         return metrics_df
 
-    def predict(self, df: pd.DataFrame, decompose: bool = True, raw: bool = False):
+    def predict(self, df: pd.DataFrame, decompose: bool = True, raw: bool = False, auto_extend=False):
```
ourownstory, line 1115: To avoid the linked issues being raised, `auto_extend` should default to `True`, unless my logic is off?

neuralprophet/forecaster.py
```diff
@@ -749,6 +749,7 @@ def add_events(
     def add_country_holidays(
         self,
         country_name: Union[str, list],
+        subdivision_name: Optional[Union[str, dict]] = None,
```
ourownstory, line 752: seems like the subdivision commits slipped in here. 
we can either remove those commits or merge that PR first and be careful that these lines do not longer reappear due to commit squashing.

neuralprophet/configure.py
```diff
@@ -397,12 +397,12 @@ def __post_init__(self):
             self.seasonality_local_reg = False
 
         # If seasonality_local_reg = True
-        if self.seasonality_local_reg == True:
+        if self.seasonality_local_reg:
```
ourownstory, line 400: This also seems to come from another PR

ourownstory (changes requested):
I added three comments, please check

neuralprophet/configure.py
ourownstory: Also, this change will always override a set `seasonality_local_reg` and overwrite it to be `1` which does not appear intentional?

MaiBe-ctrl: No was trying to solve the flake8 errors - doesn't make any difference anyways

neuralprophet/forecaster.py
MaiBe-ctrl: auto_extend for me means that we don't do the cutting at the end. So we cut all the time excpet for the case where it will cause teh whole dataframe to be nan. Can still reverse it

Issue #1565, opened 2024-04-18
Unable to extract LR coefficients


jwretham:
Hi, I'm using NP as part of a forecasting tool I'm building.

I'd really like to be able to access the linear regression coefficient estimates for my additional regressors.

In Prophet I'm able to do this using the regressor_coefficients() function, however I can't find anything in the NP documentation about how to do this.

Any help/suggestions apprecaiated

Thanks

Joe

jwretham:
Hi Constantin - so sorry, I absolutely missed that you'd responded on this. Thanks for taking the time to put this together - let me give it a try and I'll get back to you.

Joe

PR #1597, opened 2024-06-23
[minor] Add function to retrieve regressor coefficients

Constantin343:
## :microscope: Background

- Make regressor coefficients accessible via a function https://github.com/ourownstory/neural_prophet/issues/1565

## :crystal_ball: Key changes

- Added a new function to TimeNet Class which helps the user retrieve the coefficients for future regressors, lagged regressors, and events

### DIFF
--- a/neuralprophet/time_net.py
+++ b/neuralprophet/time_net.py
@@ -5,6 +5,7 @@ from functools import reduce
 from typing import Dict, List, Optional, Union
 
 import numpy as np
+import pandas as pd
 import pytorch_lightning as pl
 import torch
 import torch.nn as nn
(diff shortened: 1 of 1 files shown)


neuralprophet/time_net.py
```diff
@@ -987,6 +988,73 @@ def denormalize(self, ts):
...
+        -------
+            pd.DataFrame: A DataFrame containing the following columns:
+                - regressor: Name of the regressor.
+                - regressor_mode: Mode of the regressor ('additive' or 'multiplicative').
+                - coef: Average coefficient value for the regressor.
+        """
+        coefficients = []
+
+        # Helper function to calculate the average weight
+        def calculate_average_weight(weight):
```
ourownstory, line 1009: This function applied to both future-reg/events and lagged-reg/AR will be hard to implement correctly. I suggest splitting it into two - one function for model components where each named input/regressor is a single value (future-reg/events) and one for components where an input/regressor consists of multiple values (lagged-reg/AR). The first should return a single coefficient per regressor, while the latter returns multiple coefficients (one per number of lags inputed).

neuralprophet/time_net.py
```diff
@@ -987,6 +988,73 @@ def denormalize(self, ts):
...
+                return weight.mean(axis=0).mean(axis=0)
+            elif weight.ndim == 1:
+                return weight.mean(axis=0)
+            else:
+                return weight
+
+        # Future Regressors
+        if self.config_regressors is not None and self.config_regressors.regressors is not None:
+            for name, config in self.config_regressors.regressors.items():
+                regressor_component = self.future_regressors
```
ourownstory, line 1020: Note: Future regressors can be a single value, or a NN

neuralprophet/time_net.py
```diff
@@ -987,6 +988,73 @@ def denormalize(self, ts):
...
+                if config.mode == "additive" and "additive" in regressor_component.regressor_params:
+                    coef = regressor_component.regressor_params["additive"].data.cpu().numpy()
+                elif config.mode == "multiplicative" and "multiplicative" in regressor_component.regressor_params:
+                    coef = regressor_component.regressor_params["multiplicative"].data.cpu().numpy()
+                else:
+                    coef = [None]
```
ourownstory, line 1026: Are there cases where this should end up being `None`?

neuralprophet/time_net.py
ourownstory: Note: to achieve the described outcomes, you may also need to adjust the code below, not just `calculate_average_weight`

ourownstory (changes requested):
Thank you Constantin!

It is looking good overall! 

You are headed in the right direction - I added some comments on how you can address some issues regarding the coefficient computation. 

Please also add one or more tests covering your new functionality, a micro example in the docstring, and add a short feature-guide notebook demonstrating the use of your method.

neuralprophet/time_net.py
Constantin343: Yes makes sense! I separated it into two functions and also now return the corresponding lag for each coefficient to improve the usability.

neuralprophet/time_net.py
Constantin343: No, that should not be the case.

neuralprophet/time_net.py
Constantin343: Thanks for the comment! I adapted the function to also support NN for future regressors.

neuralprophet/time_net.py
```diff
@@ -989,6 +990,98 @@ def denormalize(self, ts):
     def train_dataloader(self):
         return self.train_loader
 
+    def get_future_and_event_regressor_coefficients(self):
```
ourownstory, line 993: What do you think about splitting future-reg and events into two methods?

neuralprophet/time_net.py
```diff
@@ -989,6 +990,98 @@ def denormalize(self, ts):
...
+                if event_config.mode == "additive" and "additive" in self.event_params:
+                    coef = self.event_params["additive"].data.cpu().numpy()
+                elif event_config.mode == "multiplicative" and "multiplicative" in self.event_params:
+                    coef = self.event_params["multiplicative"].data.cpu().numpy()
+                coef_avg = calculate_average_weight(coef)
+                coefficients.append({"regressor": event, "regressor_mode": event_config.mode, "coef": coef_avg})
+
+        return pd.DataFrame(coefficients)
+
+    def get_lagged_regressor_coefficients(self):
```
ourownstory, line 1061: Do you mind duplicating this for AR?

ourownstory (changes requested):
Thank you for upgrading the coefficient computation!

I left few small comments in the review. Beyond, Do you mind adding:

- Please also add one or more tests covering your new functionality, 
- a micro example in the docstring, 
- and add a short feature-guide notebook demonstrating the use of your method (both directly, and indirectly via the plotting methods).
- Please also look into the existing plotting code and make sure that that code uses your code for visualizing coefficients in plots. Please also make sure that if your code yields different results, that those differences are intentional/meaningful.

Constantin343:
Thanks for the feedback! I wasn't aware that some of the weight calculations were already available and used in the plotting function. I adapted the functions to integrate cleaner with the existing code. 
The existing plot function didn't seem to work with NN future regressors, so I am using `get_future_regressor_coefficients()` there now. For the other coefficient calculations I am also using some of the existing code to avoid duplicate functionality.

PR #1600, opened 2024-06-26
[Minor] Make tests deterministic

MaiBe-ctrl:
-  Add flag to make tests deterministic to better track the metrics and performance of the fitted models.


ourownstory (commented):
Great to see the tests be deterministic now!

I looked through the code and have 3 thoughts:

- What do you think about moving the `deterministic` flag from the `NeuralProphet` class init to the `train` method? AFAIK it only affects training?
- Can you please update the tutorial on reproducibility to include this new flag?
- Looks like many of the 'glocal' seasonality tests are removed - is this intentional?

MaiBe-ctrl:
- Yes, actually it makes more sense to have the flag in the `train` rather than the `model` itself. I will update it.
- For some reason, some of the tests in `glocal` are duplicated. That's also the reason for the linting issues.

ourownstory (approved):
Thank you!

PR #1608, opened 2024-07-03
[Minor] Torchify TimeDataset

MaiBe-ctrl:
## :microscope: Background

- Speedup the TimeDataset class 

### DIFF
--- a/neuralprophet/time_dataset.py
+++ b/neuralprophet/time_dataset.py
@@ -96,24 +96,37 @@ class TimeDataset(Dataset):
             self.config_regressors
         )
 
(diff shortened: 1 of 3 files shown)


ourownstory:
Note - In final review, make sure that ignoring the ID column and "__df__" in TimeDataset has no issues wrt global/local modeling.

MaiBe-ctrl:
- PyTorch can only support tensors of numerical or boolean values -> The casting at the `init` function is very important, however we have to watch out with the exact types (int, float) and the precision (32, 64), otherwise this might lead to problems.

ourownstory (approved):
Great work Maisa!!

Issue #1548, opened 2024-02-23
[Devops] Separate plotting dependencies from main library


ourownstory:
- Separate plotting dependencies, exclude them if extras "minimal" is set.
- Separate all plotting tests and run separate job with the correct dependencies. Run non-plotting tests without plotting dependencies.

PR #1611, opened 2024-07-05
[Major] Separate plotting dependencies to enable running without plotting

kneureither:
## :microscope: Background

Poetry can now be installed in a minimal mode by using `poetry install --only main`. Minimal mode does only include the main dependencies, specified under `[tool.poetry.dependencies]`. Matplotlib could also be excluded in the future, by moving it into the `[tool.poetry.group.plotting.dependencies]` in `pyproject.toml`.
All tests are passed, except `test_model_performance.py` which saves svg files. 
fixes issue #1548.

## :crystal_ball: Key changes

- Add a poetry group for plotting (for now it only includes plotly, matplotlib could also be added in the same manner)
- Catch plotly import in files (`forecaster.py`, `plot_forecast_plotly.py`, `plot_model_parameters_plotly.py`) and display warning, but proceed without plotting anything.
- status of plotly import is handled by `plotting_backend` which is set to `"no-backend-installed"`

### DIFF
--- a/neuralprophet/plot_forecast_plotly.py
+++ b/neuralprophet/plot_forecast_plotly.py
@@ -2,7 +2,6 @@ import logging
 
 import numpy as np
 import pandas as pd
-import warnings
 
 try:
     import plotly.express as px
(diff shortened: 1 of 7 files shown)


kneureither:
Can you give some feedback on the approach @ourownstory? Is that how you envisioned this to work? I would add further libraries to the plotting group and deal with them in the same manner as with plotly. What do you think about using the `plotting_backend` variable to check whether the corresponding modules are installed?

PR #1630, opened 2024-08-13
[Major] simplify learning rate finder and use log-mean of default and smooth suggestion

ourownstory:
The LR finder implementation had changed with the transition to using the Lightning implementation. This PR simplifies that implementation and additionally uses the logarithmic mean value of our smooth and the default version.


ourownstory (commented):
yaiy

PR #1700, opened 2025-10-28
modifications to accomodate numpy 2

norinasunfospha:
## :microscope: Background

I want to use neuralprophet with numpy 2 since my other packages require it.

## :crystal_ball: Key changes

I'm updating the package to use numpy 2


MikeB2019x:
I think when I encountered this the problem was that `process.py` was using `np.NaN`.  I converted all those instances (I think there are 4) to `np.nan` to comply with numpy 2.0+.  That fixed it all.

